PR #1, opened 2020-03-17
base setup

ts-nitish-a-jain:
This PR does the groundwork for the development to start, includes:

- Project barebones
- Build
- Linting and formatting
- Unit-testing & coverage
- API Docs
- Licensing & Publishing
- CI/CD setup
- Some Github meta-config


package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "js-miniapp",
+  "version": "0.1.0",
+  "description": "Mini App SDK for JavaScript",
+  "main": "index.js",
```
corycaywood: This should point to `build/index.js`.

ts-nitish-a-jain:
> It's looking good 👍
> 
> Should we also add WebPack? We can manage the build through NPM scripts instead, but it might started to get complicated.
> 
> I think we also need to add Babel. It looks like the `gts` setup is targeting `es2018` which might not support the browsers we need. Also, babel lets us explicitly specify the platforms we want to target, i.e. `'browsers': ['> 1%', 'last 4 versions', 'iOS 11.0', 'Android 4.1']`

@corycaywood It is quite a discussed situation it seems 🤔 For our purposes, `tsc` should suffice. 
I was further having a look at the articles, there wasn't a substantial reason supporting babel in our situation of now. Maybe in the future, we could add it, if the situation really arises. What do you think? 
In the meanwhile, the simple and focused setup would help to speed up the onboarding while gts is overridden.

https://blog.logrocket.com/choosing-between-babel-and-typescript-4ed1ad563e41/
https://iamturns.com/typescript-babel/

package.json
ts-nitish-a-jain: Noted. Thanks.

package.json
```diff
@@ -0,0 +1,41 @@
...
+  "dependencies": {},
+  "devDependencies": {
+    "@types/chai": "^4.2.11",
+    "@types/mocha": "^7.0.2",
+    "@types/node": "^10.0.3",
+    "chai": "^4.2.0",
+    "gts": "^1.1.2",
```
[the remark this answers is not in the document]
ts-nitish-a-jain: Glad that it adds up 😄

.circleci/config.yml
```diff
@@ -0,0 +1,103 @@
...
+jobs: # a collection of steps
+  build: # runs not using Workflows must have a `build` job as entry point
+    working_directory: ~/mini-js-sdk # directory where steps will run
+    docker: # run the steps with Docker
+      - image: circleci/node:10.16.3
+    steps:
+      - checkout
+      - run:
+          name: update-npm
+          command: 'sudo npm install -g npm@latest'
```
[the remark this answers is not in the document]
ts-nitish-a-jain: Well, I was trusting the Circle CI team for this one 😄( referring to the documentation). I guess we can remove it for increased stability. They might have had different reasons for promoting it.

.circleci/config.yml
```diff
@@ -0,0 +1,103 @@
...
+            npm publish 
+      - run:
+          name: Publish Documentation
+          command: |
+            set -e
+            npm run docs
+            if [[ ! $CIRCLE_TAG == *"-"* ]]; then
+              git checkout gh-pages
+              cp -R docs/. ./docs
+              git add docs _versions
```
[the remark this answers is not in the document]
ts-nitish-a-jain: Thanks for checking in the `_versions` part. Noted.

The typedoc website content generation is pretty good (verbose with more controls). I was thinking of hosting that on GH instead of markdown (which is pretty plain and basic). Surely, in a follow-up PR though.

package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "js-miniapp",
+  "version": "0.1.0",
+  "description": "Mini App SDK for JavaScript",
+  "main": "build/index.js",
```
suprj: I ran `npm run compile` to generate the compiled JS files and the index file was located at `build/src/index.js`. I don't know which we'd want to shoot for, but I thought you should know.

suprj:
This looks great (this is RJ)!

One feedback that I have is that if we run `npm publish` (in Circle CI), I believe we won't have the `build` directory in the package since npm ignore files in `.gitignore` (see [this](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package)). We can probably add an `.npmignore` file to have more control over which files and directory we'd want to include in the published package.

package.json
corycaywood: Oh, you're right. I think this is an issue with `tsconfig`. It needs to use `./src` as the root.

Also, isn't it conventional to use `dist` as the distribution build folder? Maybe we should use that instead of `build`?

suprj: Yup, setting `src` as the root directory makes sense. In that case, you probably have to move the test files in there too since they're written in TypeScript (I think there will be issues it they live outside the root directory).

I'm not sure what the convention is, but I've seen `dist` more than `build` lately. I think either will work as long all the pieces are follow one or the other.

corycaywood:
> One feedback that I have is that if we run `npm publish` (in Circle CI), I believe we won't have the `build` directory in the package since npm ignore files in `.gitignore` (see [this](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package)). We can probably add an `.npmignore` file to have more control over which files and directory we'd want to include in the published package.

That's a good point about npm ignore files. We should add either an `.npmignore` file or a whitelist of `files` in package.json.

Regarding the `build` folder on CircleCI, the build folder is saved to the workspace and reattached during the `publish` step so the build files should theoretically still be there. However in practice I've had trouble getting the CircleCI workspaces to work correctly between jobs. So maybe it's better if we run the `compile` step again in the `publish` job? Or maybe we just need to make sure we understand that the workspace is working correctly?

package.json
ts-nitish-a-jain: guess, `build` comes baked with `gts`(so google of them lol)

ts-nitish-a-jain:
> This looks great (this is RJ)!
> 
> One feedback that I have is that if we run `npm publish` (in Circle CI), I believe we won't have the `build` directory in the package since npm ignores files in `.gitignore` (see [this](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package)). We can probably add an `.npmignore` file to have more control over which files and directories we'd want to include in the published package.

Thanks, @suprj for sharing your feedback, I have updated some. See if it is better now. 😄 Thanks.!

ts-nitish-a-jain:
> > One feedback that I have is that if we run `npm publish` (in Circle CI), I believe we won't have the `build` directory in the package since npm ignore files in `.gitignore` (see [this](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package)). We can probably add an `.npmignore` file to have more control over which files and directory we'd want to include in the published package.
> 
> That's a good point about npm ignore files. We should add either an `.npmignore` file or a whitelist of `files` in package.json.
> 
> Regarding the `build` folder on CircleCI, the build folder is saved to the workspace and reattached during the `publish` step so the build files should theoretically still be there. However in practice I've had trouble getting the CircleCI workspaces to work correctly between jobs. So maybe it's better if we run the `compile` step again in the `publish` job? Or maybe we just need to make sure we understand that the workspace is working correctly?

@corycaywood thanks..! The attach should work. If our compile times are short over the period of time, we could transcend sooner in the near future.

tslint.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "gts/tslint.json",
+  "linterOptions": {
+    "exclude": [
+      "**/*.json"
+    ]
+  }
+}
```
minh-rakuten: I cannot add an excluding rule when run `gts check`.
Here is my snippet code for exception lint:
`"rules": {
    "no-any": false
  }`

Could you try it to see if we can add exception lint with gts?

ts-nitish-a-jain: @minh-rakuten I think the base rules look fine. Let's check if we need to customize and the background behind. + Let's do it beyond the scope of this PR.

package.json
```diff
@@ -0,0 +1,61 @@
...
+    "check": "gts check",
+    "clean": "gts clean",
+    "compile": "tsc -p .",
+    "prebuild": "npm run clean && npm run check",
+    "build": "npm run clean && npm run check && tsc -p . && npm run test",
+    "fix": "gts fix",
+    "test": "mocha -r ts-node/register tests/**/*.spec.ts --reporter mocha-junit-reporter --reporter-options mochaFile=./test-results/mocha/results.xml",
+    "coverage": "nyc npm test",
+    "version": "npx -c 'echo $npm_package_version'",
+    "docs": "chmod +x scripts/docs.sh && scripts/docs.sh"
```
corycaywood: You should be able to commit the `docs.sh` script as executable and then you don't need to have the `chmod` here.

PR #2, opened 2020-03-27
Feature/interface_uniqueID

minh-rakuten:
# Description
- Create TypeScript and commonjs module for using the `MiniApp.getUniqueId()` feature.
- Sample usage in `sample.ts`. In order to use sample in web application, please import `bundle.ts` which generated from `npm run build`.

## Links
- [MINI-618](https://jira.rakuten-it.com/jira/browse/MINI-618)
- Previous implementation for ES6 module and Android / iOS separated interface [link](https://github.com/minh-rakuten/js-miniapp/commit/24f341cbf6c9f6e7c4d715ee1ec12df0cccdd725)

# Checklist


_config.yml
```diff
@@ -0,0 +1,14 @@
+markdown: kramdown
```
corycaywood: This is for Github Pages? This should go on the `gh-pages` branch instead of `master`.

src/miniapp.ts
```diff
@@ -0,0 +1,5 @@
+export declare namespace MiniApp {
+  export function getUniqueId(): string;
```
corycaywood: As discussed, we will need the functions in the JS SDK to be asynchronous due to iOS not supporting synchronously returning values. So the return value for this should be something like `Promise<string>`.

src/sample.ts
```diff
@@ -0,0 +1,4 @@
+// tslint:disable-next-line:variable-name
+const MiniApp = require('./miniapp');
+
+console.log(`Unique Id: ${MiniApp.getUniqueId()}`);
```
corycaywood: What is this sample for? Just for demonstrating how to use the SDK? I think that once we have the sample Mini App it will act as the sample of how to use the. SDK, so maybe we don't need this file. What do you think?

package.json
```diff
@@ -28,7 +28,8 @@
     "clean": "gts clean",
     "compile": "tsc -p .",
     "prebuild": "npm run clean && npm run check",
-    "build": "npm run clean && npm run check && tsc -p . && npm run test",
+    "buildSampleBundle": "browserify build/src/sample.js > build/src/bundle.js",
```
corycaywood: Looks like you didn't include `browserify` in the `package.json`.

src/miniapp.ts
corycaywood: Also, right now this is just an interface definition. Shouldn't this be a function something like you had before? `(window as any).MiniApp.getUniqueId()`

src/miniapp.ts
```diff
@@ -0,0 +1,5 @@
+export declare namespace MiniApp {
```
corycaywood: I think I had mentioned using `namespace`, but I think maybe it doesn't need to use it.

src/sample.ts
minh-rakuten: Agree. Let's me remove those relevant sample files.

src/miniapp.ts
minh-rakuten: Both android and ios share the same interface name-calling so we don't need it anymore. Previously, the ios calling is different `(window as any).webkit.messageHandlers.MiniApp`

docs/index.md
```diff
@@ -0,0 +1,3 @@
+{% for version in site.versions %}
```
ts-nitish-a-jain: This wouldn't be required in `master` branch. It is useful in `gh-pages` actually.

index.md
```diff
@@ -0,0 +1,3 @@
+---
```
ts-nitish-a-jain: Same for this file.

package.json
```diff
@@ -28,7 +28,7 @@
     "clean": "gts clean",
     "compile": "tsc -p .",
     "prebuild": "npm run clean && npm run check",
-    "build": "npm run clean && npm run check && tsc -p . && npm run test",
+    "build": "npm run prebuild && tsc -p . && npm run test",
```
ts-nitish-a-jain: Nice catch. 😄 
Actually, in the `build` it wouldn't require `npm run prebuild` either, as the build system will auto-trigger the `pre` step.

src/miniapp.ts
ts-nitish-a-jain: This syntax is easier to read: https://www.typescriptlang.org/docs/handbook/namespaces.html#namespaced-validators
What do you think?

minh-rakuten: I don't see much different 😃

ts-nitish-a-jain: I see that we are exporting in 3 ways. Let's document it?

ts-nitish-a-jain: Generally, declarations are exported. So a usage along namespace looks 🤔 
https://www.typescriptlang.org/docs/handbook/modules.html#export

minh-rakuten: Sure. Those exports are required to use it as a commonjs module.

minh-rakuten: I see. Let's me remove it if it's redundant.

minh-rakuten: The export of function is still required because we use it as an interface function (using the implementation later from native sdk).

minh-rakuten:
@corycaywood @ts-leojoseph-r I have updated the typescript which corresponds with bridge.js. Testing it working well with our example webapp.

package.json
```diff
@@ -28,7 +30,7 @@
     "clean": "gts clean",
     "compile": "tsc -p .",
     "prebuild": "npm run clean && npm run check",
-    "build": "npm run clean && npm run check && tsc -p . && npm run test",
+    "build": "tsc -p . && npm run test",
```
ts-nitish-a-jain, line 33: 👍

js-miniapp-sdk/src/miniapp.ts
```diff
@@ -0,0 +1,16 @@
...
+}
+
+/* tslint:disable:no-any */
+export class MiniApp implements MiniAppInterface {
+  getUniqueId(): Promise<string> {
+    return (window as any).MiniAppBridge.getUniqueId();
+  }
+}
+
+module.exports = MiniApp;
```
corycaywood, line 16: I think this should be `export default MiniApp`.

package-lock.json
```diff
@@ -0,0 +1,5389 @@
+{
```
corycaywood, line 1: We shouldn't have this file committed anymore. We should always use `yarn` which uses `yarn.lock` instead of `package-lock.json`.

js-miniapp-sample/package.json
```diff
@@ -3,7 +3,9 @@
     "version": "0.1.0",
     "description": "Mini App Sample for JavaScript",
     "main": "build/src/index.js",
-    "dependencies": {},
+    "dependencies": {
+      "browserify": "^16.5.1"
```
corycaywood, line 7: Can probably remove this one.

js-miniapp-sdk/package.json
```diff
@@ -4,7 +4,9 @@
   "description": "Mini App SDK for JavaScript",
   "main": "build/src/index.js",
   "types": "build/src/index.d.ts",
-  "dependencies": {},
+  "dependencies": {
+    "browserify": "^16.5.1"
```
corycaywood, line 8: This one isn't being used yet, right? Maybe remove the dependency for now and we can add it back in if we use browserify later.

js-miniapp-sdk/package.json
```diff
@@ -28,7 +30,7 @@
     "clean": "gts clean",
     "compile": "tsc -p .",
     "prebuild": "npm run clean && npm run lint",
-    "build": "npm run clean && npm run lint && tsc -p . && npm run test",
+    "buildSdk": "tsc -p . && npm run test",
```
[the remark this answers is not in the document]
minh-rakuten: @corycaywood when I execute `yarn sdk build`, it did not go into this command in `js-miniapp-sdk`. That's strange so I rename it into `buildSdk`.

js-miniapp-sdk/src/miniapp.ts
minh-rakuten: `MiniApp` here is the class, not the instance, so the correct one is `module.exports =`

corycaywood: `module.exports` is the es6 style of exporting. Let's use the TypeScript style instead. The compiler will ensure that it's compatible with es5/es6.

js-miniapp-sdk/package.json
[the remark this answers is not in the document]
corycaywood: Hmm, strange `build` is working for me locally. It's just that `buildSdk` is longer to type and more difficult to remember. @ts-nitish-a-jain @Climbatize @ts-leojoseph-r Can you confirm which one is working for you?

js-miniapp-sdk/src/miniapp.ts
corycaywood: Oh sorry, I got that wrong. `module.exports` is the commonjs/node way of exporting. Es6 uses the same style as Typescript (`export default`), so we should use that instead.

minh-rakuten: @corycaywood do you mean that we should change commonjs to es6?

corycaywood: Yeah, I think we should use the es6/TypeScript style. Because TypeScript will compile it to the correct form.

minh-rakuten: When I tried to build a sample app, it has no problem with commonjs style 🤔 

On my latest commit, I have converted `commonjs` module into `es6`. My sample app is working well with it but please note that it requires `esmify`, `browser-resolve` if using `browserify` in sample app.

PR #3, opened 2020-03-31
Bridging script that will be used for native platform

ts-leojoseph-r:
# Description
Javascript that will be used in the native platform that will be used as a bridge with Mini App

# Checklist


src/bridging-script/ios-bridge.js
```diff
@@ -0,0 +1,29 @@
+let MiniAppBridge = {}
```
ts-nitish-a-jain, line 1: Looks good. 
@ts-leojoseph-r Can we add the generic one which works for the platform?

ts-leojoseph-r: Sorry, I don't get it. Platform can access it using `window.MiniApp.getUniqueId()` which is generic. We created MiniAppBridge object to move the exec and execCallback methods inside an object.

corycaywood: Maybe this should be defined as `window.MiniAppBridge`?

corycaywood: Regarding having generic JS for both platforms: we had discussed having three files, something like `common-bridge.js`, `ios-bridge.js`, and `android-bridge.js`. However, this will require us to bundle the JS before it can be used on the native side. That's a little tricky because then we will need to have the bundled JS committed to the repo.

I guess we could setup the CI job so that it will commit the bundled JS every time something is merged to master. That would also allow us to use TypeScript and ensure that the bundled JS is targeting ES5.

Here's my suggestion:
1. For now, keep what we have here and don't use bundling.
2. Rename this file to just `bridge.js`, and then for the Android implementation we can add platform detection for knowing whether to use `MiniApp.postMessage` or `webkit.messageHandlers.MiniApp.postMessage`.
3. In the future, we can add bundling, TypeScript, and split into separate files. Also will setup CI to commit the bundled JS when merging to master.

This will allow us to be unblocked on the native side for the moment. What do you think?

src/bridging-script/ios-bridge.js
```diff
@@ -0,0 +1,29 @@
...
+var queue = []
+
+MiniAppBridge.exec = function(action, onSuccess, onError) {
+    const callback = {}
+    callback.onSuccess = onSuccess;
+    callback.onError = onError;
+    callback.id = Math.random();
```
corycaywood, line 8: Could we just use a sequential ID here? i.e. something like `callback.id = MiniAppBridge.callbackId++`

src/bridging-script/ios-bridge.js
```diff
@@ -0,0 +1,29 @@
...
+    queue.unshift(callback)
+    webkit.messageHandlers.MiniApp.postMessage(JSON.stringify({action: action, id: callback.id}));
+}
+
+MiniAppBridge.execCallback = function(messageId, value) {
+    var queueObj = queue.find(callback => callback.id = messageId)
+    queueObj.onSuccess(value);
+    queue.shift(queueObj)
```
corycaywood, line 16: I think this isn't correct, is it? `shift` can only remove the first element, it doesn't accept parameters: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/shift

I think this needs to use `slice` maybe?

src/bridging-script/ios-bridge.js
ts-leojoseph-r: We can also inject the JS which was hosted remotely. If we get to bundle the generic JS and host it and we can use the same in the native.
But for now, I agree with the suggestions and I will make changes accordingly

src/bridging-script/ios-bridge.js
ts-leojoseph-r: Looks like slice returns a new array object, so I have used `splice()`

src/bridging-script/ios-bridge.js
ts-leojoseph-r: In that case, we would be needing an extra global variable for this. Is that ok?

src/bridging-script/ios-bridge.js
ts-leojoseph-r: > Maybe this should be defined as `window.MiniAppBridge`?
I don't feel good about attaching the `MiniAppBridge` object to the window variable. As we are already having `window.MiniApp` in window object. What do you think ?

bridge/bridge.js
```diff
@@ -0,0 +1,44 @@
...
+MiniAppBridge.execCallback = function(messageId, value) {
+    var queueObj = queue.find(callback => callback.id = messageId)
+    queueObj.onSuccess(value);
+    var messageObjIndex = queue.indexOf(queueObj)
+    if(messageObjIndex != -1) {
+        queue.splice(messageObjIndex, 1);
+    }
+}
+
+window.MiniApp = {
```
minh-rakuten, line 34: Actually, we don't need `window.` here. Just `MiniApp` is enough for code running and it's also on the same page with TypeScript [interface](https://github.com/rakutentech/js-miniapp/blob/c0200d459403dfd2c38db87976ab91a342c307ba/src/miniapp.ts)

bridge/bridge.js
```diff
@@ -0,0 +1,44 @@
+let MiniAppBridge = {}
```
corycaywood, line 1: This will be called from the native side, so we need it to be scoped globally (`let` is for local scoping). This isn't going to be accessible to the calls from the native side (at least not for Android).

If you're worried about polluting the JS side with more objects, you could set it on the `MiniApp` object. i.e.
```
window.MiniApp = {};
MiniApp.bridge = MiniAppBridge;
MiniApp.getUniqueId = ....
```

bridge/bridge.js
corycaywood: Actually, I think we should be declaring `MiniApp` in the JS SDK, and then the calls in the JS SDK will basically be just wrapping calls to this one. So maybe this injected code should only declare a `MiniAppBridge` object?
```
window.MiniAppBridge = {};
MiniAppBridge.exec = ...
MiniAppBridge.execCallback = ...
MiniApp.getUniqueId = ....
```

And then in the JS SDK, the code will be something like:
```
window.MiniApp = {
   getUniqueId: () => MiniAppBridge.getUniqueId()
```

Does that make sense? What do you think?

bridge/bridge.js
corycaywood: Sorry, I had second thoughts on this. See my comment below

bridge/bridge.js
ts-leojoseph-r: Changes are done.  Now the JS SDK can access like below,

```
window.MiniApp = {
   getUniqueId: () => MiniAppBridge.getUniqueId()
}
```

corycaywood:
I think this is looking good 👍 

I've merged the changes for making this a mono repo, so you will need to rebase on master. The bridging code should be moved to the `js-miniapp-bridge` folder.

js-miniapp-bridge/bridge.js
```diff
@@ -0,0 +1,42 @@
...
+    iOS: function() {
+        return navigator.userAgent.match(/iPhone|iPad|iPod/i);
+    }
+};
+
+MiniAppBridge.exec = function(action, onSuccess, onError) {
+    const callback = {}
+    callback.onSuccess = onSuccess;
+    callback.onError = onError;
+    callback.id = Math.random();
```
ts-nitish-a-jain, line 16: How about using UUIDs instead?

js-miniapp-bridge/bridge.js
```diff
@@ -0,0 +1,42 @@
...
+var isPlatform = {
+    Android: function() {
+        return navigator.userAgent.match(/Android/i);
+    },
+    iOS: function() {
+        return navigator.userAgent.match(/iPhone|iPad|iPod/i);
+    }
+};
+
+MiniAppBridge.exec = function(action, onSuccess, onError) {
```
ts-nitish-a-jain, line 12: As we were talking, let's put some docs as well, please.

js-miniapp-bridge/bridge.js
corycaywood: I don't think JavaScript has a built in method for generating a UUID. I mentioned using just a sequential ID. So we can have some sort of `next` method that returns `id++`. We just need each ID to be unique during the session.

js-miniapp-bridge/bridge.js
ts-leojoseph-r: Done

js-miniapp-bridge/bridge.js
ts-leojoseph-r: Yes, JS doesn't have a built-in method for generating a UUID. Updated code to have sequential unique ID now

js-miniapp-bridge/bridge.js
```diff
@@ -0,0 +1,64 @@
...
+    const callback = {}
+    callback.onSuccess = onSuccess;
+    callback.onError = onError;
+    callback.id = ++uniqueId;
+    messageQueue.unshift(callback)
+    if(isPlatform.iOS()){
+        webkit.messageHandlers.MiniAppiOS.postMessage(JSON.stringify({action: action, id: callback.id}));
+    } else {
+        window.MiniAppAndroid.getUniqueId(JSON.stringify({action: action, id: callback.id}))
```
minh-rakuten, line 29: Thanks from our discussion. This can be similar to iOS:
`window.MiniAppAndroid.postMessage(JSON.stringify({action: action, id: callback.id}))`

ts-leojoseph-r: Done

PR #4, opened 2020-04-06
refactor: Organize repo as Mono Repo using Yarn Workspaces

corycaywood:
# Description
This also means that Yarn should be used for installing dependencies and running scripts. I have added some additional documentation to the readme about this.

# Checklist


js-miniapp-sdk/package.json
```diff
@@ -0,0 +1,61 @@
+{
+  "name": "js-miniapp-sdk",
+  "version": "0.1.0",
+  "description": "Mini App SDK for JavaScript",
+  "main": "build/src/index.js",
+  "types": "build/src/index.d.ts",
+  "dependencies": {},
```
ts-nitish-a-jain, line 7: How about adding?
```
"dependencies": {
    "cross-env": "5.0.5"
  }
```

corycaywood: I think that was just an example on the Yarn documentation, so we don't need it. We can add it in the future if we need to set any environment variables.

PR #6, opened 2020-04-17
Feature/feature/MINIBUILD-58_MOVING_POC_FROM_MAB_TO_MAP_REPO

Rajesh-Sundaram:
# Description
Replaced js-miniapp-sample workspace with MAB POC source
updated .flowconfig to detect yarn dependencies
Added CI Job for building sample sub-project.

## Links
[MINIBUILD-60](https://jira.rakuten-it.com/jira/browse/MINIBUILD-60)

# Checklist
- [X] I ran `yarn install` without issues
- [X] I ran `yarn sample start` without issues
- [X] I ran `yarn sample test` without issues
- [X] I ran `yarn sdk build` without issues


js-miniapp-sample/README.md
```diff
@@ -0,0 +1,133 @@
+# miniapp-build-common
```
suprj, line 1: I would recommend changing this to `js-miniapp-sample` to reflect the new root folder

js-miniapp-sample/README.md
```diff
@@ -0,0 +1,133 @@
...
+  - [Code Formatting](#code-formatting)
+  - [Naming Convention](#naming-convention)
+    - [Branches](#branches)
+    - [Commits](#commits)
+  - [Branching policy](#branching-policy)
+  - [Others](#others)
+
+## Requirements
+
+- If you're a new member **[please read the MiniAppBuild guidelines](https://confluence.rakuten-it.com/confluence/display/ONEAPP/MAB+-+On-boarding+Guide+for+Engineers)**
```
suprj, line 28: Not sure if there are any security concerns with exposing our internal Confluence URLs. Also, since the repo is moved to a different host, I believe the below requirement about the SSH keys is no longer needed.

js-miniapp-sample/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Code Formatting
+
+- The following dev-dependencies have been installed for this purpose: `husky, lint-staged, prettier`
+- Whenever we make a commit Prettier will format the changed files automatically.
+- You might want to integrate Prettier in your favorite editor as well. Please check the [Prettier Editor Integration](https://prettier.io/docs/en/editors.html).
+
+## Naming Convention
+
+### Branches
```
suprj, line 103: Since the repo is now in a mono repo, I'm not sure if the below branching and commit guidelines still applies. It's possible that we may need to follow the guidelines set by the platform team.

js-miniapp-sample/package.json
```diff
@@ -1,25 +1,77 @@
...
-    "bugs": {
-      "url": "https://github.com/rakutentech/js-miniapp/issues"
-    },
-    "homepage": "https://github.com/rakutentech/js-miniapp#readme"
-  }
-  
\ No newline at end of file
+  "name": "js-miniapp-sample",
+  "version": "0.1.0",
+  "description": "Mini App Sample for JavaScript",
```
suprj, line 4: I would add `SDK` at the end so that the description is more clear (i.e. `Mini App Sample for JavaScript SDK`)

js-miniapp-sample/package.json
```diff
@@ -1,25 +1,77 @@
...
+    "prettify": "prettier --check src/",
+    "lint": "eslint src/ --fix-dry-run",
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "dockerBuild": "cd ci && make build",
+    "test": "react-scripts test",
+    "eject": "react-scripts eject",
+    "flow": "flow"
+  },
+  "eslintConfig": {
```
suprj, line 27: Since we have a config file for ESLint, I believe this object can be removed

js-miniapp-sample/src/App.js
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+
+import "./App.css";
```
suprj, line 3: I think this CSS file can be removed since it's not used (there are no HTML elements with an `.App` class)

Rajesh-Sundaram: https://github.com/rakutentech/js-miniapp/blob/34389703dae56344d2824442053dc4f9b19c832e/js-miniapp-sample/src/App.js#L20

updated it to use makeStyles css style

js-miniapp-sample/src/services/uuid/reducers.js
```diff
@@ -0,0 +1,25 @@
...
+
+const defaultState: UUIDState = {
+  uuid: undefined,
+};
+
+const UUIDReducer = (
+  state: UUIDState = defaultState,
+  action: UUIDAction = {}
+): UUIDState => {
+  if (action !== undefined) {
```
suprj, line 17: Instead of using two nested if statements, we could simplify this to:

```suggestion
  if (action !== undefined && action.type === SET_UUID) {
     return { ...defaultState, uuid: action.payload };
  }
  return state
```

js-miniapp-sample/src/services/uuid/types.js
```diff
@@ -0,0 +1,2 @@
+const SET_UUID = "SET_UUID";
```
suprj, line 1: This is minor, but we can export a constant in one line:

```suggestion
export const SET_UUID = "SET_UUID";
```

js-miniapp-sample/src/services/uuid/actions.js
```diff
@@ -0,0 +1,17 @@
...
+import { SET_UUID } from "./types";
+
+type GetUUIDAction = { type: String, payload: string };
+
+type UUIDAction = GetUUIDAction;
+
+const setUUID = (): GetUUIDAction => {
+  return {
+    type: SET_UUID,
+    payload: getUUIDFromMobileSdk(),
```
suprj, line 12: Getting a UUID is going to be a side effect (the definition is loose, but I understand it as any observable interaction with the outside world).

Redux follow functional programming concepts. For example, reducers are pure functions. Since actions are inputs into reducers (that describe _what_ changed), they need to be pure as well. Since we're introducing an impure calculation, we'll need to update this action creator to be a thunk that returns a function, which is a common technique for delaying the evaluation until it's called.

I suggest changing the `setUUID` function to:

```javascript
const setUUID = () => {
  return dispatch => {
    dispatch({
      type: SET_UUID,
      payload: getUUIDFromMobileSdk(),
    });
  };
};
```

And then installing `redux-thunk` and setting it up as a middleware to handle thunks in `src/services/store.js`:

```javascript
import {createStore, applyMiddleware} from "redux";
import thunkMiddleware from "redux-thunk";

import reducers from "./reducers";

export default createStore(reducers, applyMiddleware(thunkMiddleware));
```


More info: https://redux.js.org/advanced/async-actions#async-action-creators

js-miniapp-sample/src/services/chatbot/reducers.js
```diff
@@ -0,0 +1,31 @@
...
+    {
+      id: 3,
+      name: "K-Chatbot",
+    },
+  ],
+};
+
+export default (state: ChatBotState = chatbotState, action: ChatBotAction) => {
+  if (action.type === SET_CHATBOTS) {
+    state = { ...state, bots: action.payload };
```
suprj, line 28: We should return the new state instead of reassigning the old one:

```suggestion
    return { ...state, bots: action.payload };
```

js-miniapp-sample/src/services/home/reducers.js
```diff
@@ -0,0 +1,23 @@
...
+
+const homeState: HomePageState = {
+  title: "POC",
+};
+
+export default (
+  state: HomePageState = homeState,
+  action: HomeAction
+): HomePageState => {
+  if (action !== undefined) {
```
suprj, line 17: Same comment here with simplifying the nested if statements

js-miniapp-sample/src/components/ToolBar.js
```diff
@@ -0,0 +1,108 @@
...
+  showDrawer: Boolean,
+  actions: any,
+  onShrinkToggle: Function,
+  onDrawerToggle: Function,
+  navItems: any,
+};
+
+const ToolBar = (props: ToolBarProps) => {
+  const classes = useStyles();
+  const [drawer, setDrawer] = useState({ show: props.showDrawer ?? false });
```
suprj, line 39: If `drawer` is an object that just has one boolean field, we could probably just use a boolean value instead for simplification:

```suggestion
  const [showDrawer, setShowDrawer] = useState(props.showDrawer ?? false);
```

js-miniapp-sample/src/components/ToolBar.js
```diff
@@ -0,0 +1,108 @@
...
+
+  useEffect(
+    () => {
+      setDrawer({ ...drawer, show: props.showDrawer });
+    }, // eslint-disable-next-line
+    [props.showDrawer]
+  );
+  const [shrinkState, setShrink] = useState({ shrink: false });
```
suprj, line 47: The same comment as above with this shrink state

js-miniapp-sample/src/services/uuid/types.js
Rajesh-Sundaram: as the app grows, i assume there will be few more constants & flow types will be added.

suprj: Yup definitely, and we are able to export each individual constants and types instead of bundling them all up.

Doing this:

```javascript
export const a = "a";
export const b = "b";
export const c = "c";
export const d = "d";

export type t = {...}
```

Instead of:

```javascript
const a = "a";
const b = "b";
const c = "c";
const d = "d";

type t = {...}

export {
  a,
  b,
  c,
  d
}

export type {
  t
}
```

Rajesh-Sundaram: i feel consolidating all the exported constants looks clean, when we have some types (sub types) in a file which are not shared to any other file. what do you think?

suprj: That sounds fine with me to consolidate all exports 👍

js-miniapp-sample/src/js_sdk/index.js
```diff
@@ -0,0 +1,5 @@
+import { v4 as uuidv4 } from "uuid";
+
+export const getUUIDFromMobileSdk = () => {
+  return uuidv4();
```
corycaywood, line 4: We've merged the JS SDK portion now, so could you please integrate that instead?

Because we're using Yarn workspaces, you should be able to use just `import MiniApp from "js-miniapp-sdk"`. Also, the `getUniqueId` function uses promises, so you will need to make some modifications to handle this.

Rajesh-Sundaram: @corycaywood 
will getUniqueId respond with uuid? i see no implementation on bridge?

corycaywood: @Rajesh-Sundaram The implementation is on the Android/iOS side, so you won't be able to test in the browser. We have been considering making some kind of mocking functionality for browser testing, but this hasn't been implemented yet.

For now, probably you will need to first check if it's running on a mobile device (perhaps using 'bowser'? https://lancedikson.github.io/bowser/docs/Parser.html#getPlatform) and return a mock value if it's not on mobile.

js-miniapp-sample/.flowconfig
```diff
@@ -0,0 +1,25 @@
...
+<PROJECT_ROOT>/src/serviceWorker.js
+<PROJECT_ROOT>/src/setupTests.js
+[include]
+
+[libs]
+
+[lints]
+
+[options]
+# all=true
```
xavimajoral, line 20: If we comment this line here is flow still going to check all the .js files?

Rajesh-Sundaram: updated the flowconfig

js-miniapp-sample/README.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+**Note**: to stop the local docker container run `make stop` on `Makefile` directory.
+
+**Prerequisite**: [Docker](https://docs.docker.com/) LTS
+
+## Flow
+
+- We use [Flow](https://flow.org/) as a static type checker in our App.<br/>
+- Add `// @flow` to any files you want to type check (for example, to [src/App.js](src/App.js)).<br/>
```
xavimajoral, line 74: In case Flow automatically checks all the .js files this line should be removed

js-miniapp-sample/README.md
```diff
@@ -0,0 +1,119 @@
...
+- Tech branches (mostly for technical tasks): `tech/MINIBUILD-xxxxx_short_description_of_the_task`
+- Releases branches: `release/(each project has its own version naming)`
+
+**It's really important to include the JIRA ticket ID! In case a ticket doesn't exist create one (as a tech ticket)**
+
+### Commits
+
+- Each commit should have a meaningful message. It really helps to figure out where some changes are made, even for adding comments and minor modifications.
+- Try to make the commits as small as possible and avoid several days of work worth big giant commits.
+- _Proposal_: I'd include the JIRA ticket ID on the commit message as well: `git commit -m "MINIBUILD-XXX short description of the commited task"`
```
xavimajoral, line 108: I'd remove this line and discuss later which strategy we want to follow

js-miniapp-sample/src/components/ToolBar.js
```diff
@@ -0,0 +1,107 @@
...
+    props.onDrawerToggle(showFlag);
+    setDrawer(showFlag);
+  };
+  const onDrawerShrink = () => {
+    const shrinkFlag = !shrinkState;
+    props.onShrinkToggle(shrinkFlag);
+    setShrink(shrinkFlag);
+  };
+
+  const onOpenClose = (type, event) => {
```
xavimajoral, line 58: Do we need the `type` param here?

Rajesh-Sundaram: removed it

js-miniapp-sample/src/js_sdk/index.js
Rajesh-Sundaram: added

js-miniapp-sample/README.md
```diff
@@ -0,0 +1,117 @@
...
+  - [Building](#building)
+  - [Export Dockerized App Image as TAR](#export-dockerized-app-image-as-tar)
+  - [Import Dockerized App Image from TAR](#import-dockerized-app-image-from-tar)
+  - [Flow](#flow)
+  - [Custom Environment Variables](#custom-environment-variables)
+  - [Code Formatting](#code-formatting)
+  - [Naming Convention](#naming-convention)
+    - [Branches](#branches)
+    - [Commits](#commits)
+  - [Branching policy](#branching-policy)
```
ts-nitish-a-jain, line 22: Do we require these into the sample app?

Rajesh-Sundaram:
> The build is still failing. How are we moving on this one? @corycaywood @Rajesh-Sundaram
I want to have the SDK build as the dependency in yarn workspace when building js-miniapp-sample? do you have any idea how i can achieve it without compiling sdk in **build-sample** job of  js-miniapp-sample (Ref: circleci conf)
@corycaywood @ts-nitish-a-jain 
image

minh-rakuten:
> > image
There is the missing dependency, isn't it?
I think you should declare those dependencies in `package.json` or doing sth forcing the CI downloads needed dependencies.

corycaywood:
> > The build is still failing. How are we moving on this one? @corycaywood @Rajesh-Sundaram
> > I want to have the SDK build as the dependency in yarn workspace when building js-miniapp-sample? do you have any idea how i can achieve it without compiling sdk in **build-sample** job of  js-miniapp-sample (Ref: circleci conf)
> > @corycaywood @ts-nitish-a-jain
> > image

@Rajesh-Sundaram The SDK and sample app compilation are working correctly, right? The sample app is compiling just fine. 

The CI is failing on the `yarn sample flow` step. Looks like flow doesn't support Yarn Workspaces: https://github.com/facebook/flow/issues/5183

I'm not sure the best way to handle it, but someone on that ticket recommended this: https://github.com/ImmoweltGroup/flow-mono-cli

Rajesh-Sundaram:
> > > image
> 
> @Rajesh-Sundaram There is the missing dependency, isn't it?
> I think you should declare those dependencies in `package.json` or doing sth forcing the CI downloads needed dependencies.

@minh-rakuten 
i think you are right. 

@corycaywood 
even though we do "sdk compile" post installation of all the subworkspaces, **yarn sdk compile**
command generates files inside **js-miniapp-sdk/build**, i don't know how this built can be referenced by js-miniapp-sample automatically, if its not added as dependency in "js-miniapp-sample". should i add it as dependency?

PR #8, opened 2020-04-22
SDK: Fix typescript module resolution and add unit test

corycaywood:
# Description
The tests will not run correctly when using 'es2015'. Also 'commonjs' is the expected module type when using the SDK an NPM package. Also added a unit test for `getUniqueId`.

# Checklist


js-miniapp-sdk/tsconfig.json
```diff
@@ -3,11 +3,11 @@
     "compilerOptions": {
         "rootDir": ".",
         "outDir": "build",
-        "module": "es2015",
+        "module": "commonjs",
```
minh-rakuten, line 6: Converting into `commonjs` also requires the change of syntax in `miniapp.ts` `index.ts`.
Please take a look here [commonjs -> es6](https://github.com/rakutentech/js-miniapp/commit/51027d34aff468855cb176b00a589385056112dc)

corycaywood: This setting only affects the `build` output by the TypeScript compiler. For the imports/exports in `.ts` files, we can use whichever style that we want and then compiler will convert it to the one specified in `tsconfig.json`.

minh-rakuten: If it's working with the current sample app then it's okay.
My sample app is using js in `build` folder in the end and it's not working.Perhaps my sample app went wrong itself 😢 but you need to check on those generated js files.

corycaywood: Yeah, the generated JS has to be used as an NPM module and compiled/bundled by the App which is using it (i.e. babel, browserify, webpack, etc). So it won't work if it's used as is in a Web App.

Probably we should also add browserify or something so we can have a compiled/bundled JS file as well. But it's not needed at this point for the Sample App because they are already using a bundler.

PR #12, opened 2020-05-11
(bridge) Fix for Android 5 & 6

corycaywood:
# Description
Some JS features were unsupported on older browsers. The bridge script was not working on Android 5 and 6 because the default install of Chrome is a very outdated version. This means our bridge script needs to use features which target ES5. This commit replaces arrow functions (=>) with normal functions and replaces 'Array.find' with 'Array.filter'. Also moved the 'module.exports' to an 'if' statement to avoid errors in the browser.

Also added CircleCI job for bridge and setup linting with recommended settings and using es5 to ensure we don't use unsupported features

# Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,6 +56,30 @@ jobs: # a collection of steps
             - js-miniapp-sdk/build/
             - js-miniapp-sdk/publishableDocs/
 
+  build-bridge:
+    working_directory: ~/mini-js-bridge
+    docker:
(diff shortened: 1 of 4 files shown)


corycaywood:
> LGTM.
> Google does not support android 5 & 6 anymore. If those outdated webviews cause a lot of issues further then we should consider changing the minimum support version.

Yeah, exactly. Actually I think we still need to support 6 for the time being, but probably we can consider at least dropping 5.

PR #14, opened 2020-05-14
Fixes for doc publishing from CI

corycaywood:
# Description
- Publish docs as X.X without fix version. Docs should be published as X.X so if a new fix version is published, it will replace the documentation for the previous version. However, publishing a new minor version will still create new a docs version

-  Change typedoc to 'file' mode instead of 'module'. The module mode causes issues with Github pages because it uses underscores in the filenames i.e. '_index_.html'. Github pages treats files that start with an underscore as hidden, so these pages were showing as 404.

- Fix breadcrumb parent link. Typedoc generates the parent link as 'README.md', but we are using 'index.md'. Typedoc doesn't have any options to easily change this (except hiding breadcrumbs), so I've added a 'README.md' wh
ich just redirects to index.

# Checklist

### DIFF
--- a/js-miniapp-sdk/scripts/docs.sh
+++ b/js-miniapp-sdk/scripts/docs.sh
@@ -22,3 +22,6 @@ npx typedoc --includeVersion --out $DIR_DOCS/api src --plugin typedoc-plugin-mar
 # Move userguide to docs and create correct folder structure
 mv $DIR_DOCS/api/README.md $DIR_DOCS/api/index.md
 cp README.md $DIR_DOCS/index.md
+# The typedoc plugin generates breadcrumbs which point to `README.md` as the parent page
+# So we must create this page and add a redirect to `index`
+cp ./scripts/readme-redirect.md $DIR_DOCS/api/README.md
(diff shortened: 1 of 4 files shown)


js-miniapp-sdk/scripts/docs.sh
```diff
@@ -14,6 +16,12 @@ version: "$PACKAGE_VERSION"
 ---"
 cat <<< "$contents" > "$FILE_VERSION_MD"
 
-npx typedoc --includeVersion --out $DIR_DOCS/api src --plugin typedoc-plugin-markdown --stripInternal --readme none
+# Generate docs
+npx typedoc --includeVersion --out $DIR_DOCS/api src --plugin typedoc-plugin-markdown --stripInternal --readme none --mode file
```
minh-rakuten, line 20: I'm not sure but by changing `--readme none` into `--readme index.md`, could it help you to define the parent as `index.html`?

corycaywood: I tried something like this, but it didn't seem to work. It still renames the file to `README.md` for some reason.

PR #16, opened 2020-05-20
Feature/MINIBUILD 130 unique id from sdk

RajeshSundaram:
# Description
- added UUID fetcher from SDK
- added default route to sample mini-app
- removed unused files & updated homepage title alignment

# Checklist
- [X] I wrote/updated tests for new/changed code
- [X] I verified mini-app fetches the Unique UUID inside Android/IOS emulator.
- [X] I ran `npm run build` without issues


js-miniapp-sample/src/components/ToolBar.js
```diff
@@ -20,7 +20,7 @@ const useStyles = makeStyles((theme) => ({
     marginRight: theme.spacing(2),
   },
   title: {
-    flexGrow: 1,
+    // flexGrow: 1,
```
Climbatize, line 23: can you remove commented code ?

PR #18, opened 2020-06-29
JS bridge changes to request permission from the SDK

ts-leojoseph-r:
# Description
JS Bridge changes to request any permissions for Android/iOS SDK

## Links
- [MINI-914](https://jira.rakuten-it.com/jira/browse/MINI-914)
- [MINI-915](https://jira.rakuten-it.com/jira/browse/MINI-915)

# Checklist

### DIFF
--- a/js-miniapp-bridge/src/bridge.js
+++ b/js-miniapp-bridge/src/bridge.js
@@ -15,7 +15,8 @@ var isPlatform = {
  * Method to call the native interface methods for respective platforms
  * such as iOS & Android
  * @param  {[String]} action Action command/interface name that native side need to execute
- * @param  {String} param Action command/interface name that native side need to execute
+ * @param  {Object} param Object that contains request parameter values like permissions.
+ * For eg., {permission: 'location'}
  * @param  {[Function]} onSuccess Success callback function
  * @param  {[Function]} onError Error callback function
  */
--- a/js-miniapp-bridge/src/bridge.js
+++ b/js-miniapp-bridge/src/bridge.js
@@ -100,13 +100,13 @@ MiniAppBridge.getUniqueId = function() {
 
 /**
  * Associating requestPermission function to MiniAppBridge object
- * @param {String} requestType Type of permission that is requested. For eg., location
+ * @param {String} permissionType Type of permission that is requested. For eg., location
  */
-MiniAppBridge.requestPermission = function(requestType) {
+MiniAppBridge.requestPermission = function(permissionType) {
   return new Promise(function(resolve, reject) {
     return MiniAppBridge.exec(
       "requestPermission",
-      requestType,
+      {permission: permissionType},
       function(success) { return resolve(success) },
       function (error) { return reject(error) }
     );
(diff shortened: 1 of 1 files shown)


js-miniapp-bridge/src/bridge.js
```diff
@@ -96,6 +97,21 @@ MiniAppBridge.getUniqueId = function() {
...
+
+/**
+ * Associating requestPermission function to MiniAppBridge object
+ * @param {String} requestType Type of permission that is requested. For eg., location
+ */
+MiniAppBridge.requestPermission = function(requestType) {
+  return new Promise(function(resolve, reject) {
+    return MiniAppBridge.exec(
+      "requestPermission",
+      requestType,
```
minh-rakuten, line 109: I'm thinking about changing `requestType` into `Int`. The code of `requestType` will be defined in android, ios and js sdk.
Example:
```
        const val GEOLOCATION = 1001
        const val LOCATION = 1002
```
@ts-leojoseph-r @corycaywood What do you think?

ts-leojoseph-r: We will end up managing/maintaining Int and String back and forth, which would create some confusion. I think it would be good to have a readable format

corycaywood: Yeah, @ts-leojoseph-r  is right. Unfortunately I think using Int would be difficult to manage and sync across platforms.

js-miniapp-bridge/src/bridge.js
```diff
@@ -96,6 +97,21 @@ MiniAppBridge.getUniqueId = function() {
     );
   });
 };
+
+/**
+ * Associating requestPermission function to MiniAppBridge object
+ * @param {String} requestType Type of permission that is requested. For eg., location
+ */
+MiniAppBridge.requestPermission = function(requestType) {
```
corycaywood, line 105: Maybe `requestType` should be named something like `permission` instead? `requestType` is a bit confusing because it puts emphasis on the `request` part rather than the `permissions`.

js-miniapp-bridge/src/bridge.js
corycaywood, line 109: Will it make more sense to pass a JSON object as the params? It would let the `exec` function be more generic and then we can more easily add more params in the future if needed. So here you could pass something like `{permission: 'some_permission'}`.

ts-leojoseph-r: Sure, I even thought of doing the same.

js-miniapp-bridge/src/bridge.js
```diff
@@ -15,22 +15,23 @@ var isPlatform = {
  * Method to call the native interface methods for respective platforms
  * such as iOS & Android
  * @param  {[String]} action Action command/interface name that native side need to execute
+ * @param  {String} param Action command/interface name that native side need to execute
```
corycaywood, line 18: Still need to update this comment to `Object`.

ts-leojoseph-r: Ahh, thanks.

PR #20, opened 2020-07-03
Small change to pass null object for getUniqueId interface

ts-leojoseph-r:
# Description
For getUniqueId interface, we need to pass the null object in the request parameter section

# Checklist

### DIFF
--- a/js-miniapp-bridge/src/bridge.js
+++ b/js-miniapp-bridge/src/bridge.js
@@ -93,6 +93,7 @@ MiniAppBridge.getUniqueId = function() {
   return new Promise(function(resolve, reject) {
     return MiniAppBridge.exec(
       "getUniqueId",
+      null,
       function(id) { return resolve(id) },
       function (error) { return reject(error) }
     );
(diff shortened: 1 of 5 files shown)


js-miniapp-bridge/src/bridge.js
```diff
@@ -93,6 +93,7 @@ MiniAppBridge.getUniqueId = function() {
   return new Promise(function(resolve, reject) {
     return MiniAppBridge.exec(
       "getUniqueId",
+      null,
```
ts-nitish-a-jain, line 96: Should this be `null` or an empty placeholder object, being explicit in the usage?

ts-leojoseph-r: I guess null is rational which is the representation of no value/request

corycaywood:
Looks like this change was already included in the PR from @minh-rakuten, so maybe we can close this?

ts-leojoseph-r:
Similar changes were pushed in another PR

PR #21, opened 2020-07-07
feat: (sample) Add location permissions request

corycaywood:
# Description
The SDK has added a new feature for requesting location permissions from the host app. The user is able to revoke permissions at any time, so 'requestLocationPermissions' must be called every time before using location

# Checklist


corycaywood:
> 👍 👍
> We need to let the developers aware that they need to call a permission request every time they want to access the data (permission revoke).
> Can you help me on updating the document [here](https://github.com/rakutentech/js-miniapp/blob/master/js-miniapp-sdk/README.md#3-request-permissions)?

That's a good point. I've updated the user guide with this.

PR #22, opened 2020-07-08
Update versions for alpha release

ts-nitish-a-jain:
# Description
Update versions for alpha release for gurunavi

# Checklist
- [ ] I ran `npm run build` without issues

### DIFF
--- a/js-miniapp-bridge/package.json
+++ b/js-miniapp-bridge/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-miniapp-bridge",
-  "version": "0.1.0",
+  "version": "1.1.0-alpha",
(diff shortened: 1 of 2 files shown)


js-miniapp-bridge/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "js-miniapp-bridge",
-  "version": "0.1.0",
+  "version": "1.1.0-alpha",
```
corycaywood, line 3: This one doesn't actually get released since it's manually integrated into the SDK, so the version number here doesn't affect anything. But I guess it's still good to keep them on the same version.

ts-nitish-a-jain: Exactly for the sync :)

PR #25, opened 2020-07-20
update changelog for 1.2.0 release

ts-nitish-a-jain:
# Description
Update docs for 1.2.0 release

## Links
[MINI-1254](https://jira.rakuten-it.com/jira/browse/MINI-1254?src=confmacro)

# Checklist
- [ ] I ran `npm run build` without issues

### DIFF
--- a/js-miniapp-sdk/README.md
+++ b/js-miniapp-sdk/README.md
@@ -110,6 +110,4 @@ function getId() {
 
 ## Changelog
 
-### v1.0.0 (2020-5-13)
-
-- Initial release.
+See the full [CHANGELOG](CHANGELOG.md).
--- a/js-miniapp-sdk/README.md
+++ b/js-miniapp-sdk/README.md
@@ -110,4 +110,4 @@ function getId() {
 
 ## Changelog
 
-See the full [CHANGELOG](CHANGELOG.md).
+See the full [CHANGELOG](https://github.com/rakutentech/js-miniapp/blob/master/js-miniapp-sdk/CHANGELOG.md).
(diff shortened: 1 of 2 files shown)


js-miniapp-sdk/README.md
```diff
@@ -110,6 +110,4 @@ function getId() {
 
 ## Changelog
 
-### v1.0.0 (2020-5-13)
-
-- Initial release.
+See the full [CHANGELOG](CHANGELOG.md).
```
corycaywood, line 113: I think this shouldn't be separated into a different file. The README is published to Github pages at https://rakutentech.github.io/js-miniapp/docs/1.0/ (and it also gets published to [npmjs](https://www.npmjs.com/package/js-miniapp-sdk), so the Changelog needs to be available at the same location as the user guide. 

The link here to `CHANGELOG.md` could be updated to point to the full github.com url, but I think it's not as nice as having the change log built-in to the user guide. Having it in the User Guide allows the reader to more easily search (via cmd+f) and to more easily navigate between the changelog/guide. Also, it's sometimes necessary to put links in the changelog to specific parts of the user guide, and this is easier to do if they are in the same page.

My alternative suggestion is this:
- Keep the Changelog at the bottom of the User Guide
- Add a Table of Contents at the top of the User Guide to make navigation to the Changelog easier
- Once the Changelog gets too long, we can collapse the entries for older versions so that they're expandable by clicking (via JavaScript), or we can separate old versions into a different file.
- Update the [github releases](https://github.com/rakutentech/js-miniapp/releases) with changes for each release.

Climbatize: Indeed I like your suggestion @corycaywood

ts-nitish-a-jain: As we spoke, updated the link to the absolute value of the URL.

PR #29, opened 2020-07-29
Geolocation override changes for iOS SDK

ts-leojoseph-r:
# Description
Override `navigator.geolocation.getCurrentPosition` for iOS SDK

# Checklist

### DIFF
--- a/js-miniapp-bridge/src/bridge.js
+++ b/js-miniapp-bridge/src/bridge.js
@@ -114,6 +114,24 @@ MiniAppBridge.requestPermission = function(permissionType) {
     );
   });
 };
+
+/**
+ Below code will override the navigator.geolocation.getCurrentPosition method for only iOS
+ */
+if (isPlatform.iOS()) {
+    navigator.geolocation.getCurrentPosition = function(success, error, options) {
+        return MiniAppBridge.exec(
+            "getCurrentPosition",
+            {locationOptions: options},
+            function(value) {
+                var parsedData = JSON.parse(value)
+                success(parsedData)
+            },
+            error
+        );
+    };
+}
+
 window.MiniAppBridge = MiniAppBridge;
 
 // Exported for unit testing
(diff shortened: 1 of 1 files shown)


js-miniapp-bridge/src/bridge.js
```diff
@@ -114,6 +114,24 @@ MiniAppBridge.requestPermission = function(permissionType) {
...
+ Below code will override the navigator.geolocation.getCurrentPosition method for only iOS
+ */
+if (isPlatform.iOS()) {
+    navigator.geolocation.getCurrentPosition = function(success, error, options) {
+        return MiniAppBridge.exec(
+            "getCurrentPosition",
+            {locationOptions: options},
+            function(value) {
+                var parsedData = JSON.parse(value)
+                success(parsedData)
```
corycaywood, line 128: Maybe the `JSON.parse` should be in a `try/catch`? I know that the iOS side is directly controlling what's returned here, but seems like a good idea in case of some edge case where one of the values can't be parsed correctly:

```suggestion
                try {
                    var parsedData = JSON.parse(value)
                    success(parsedData)
                } catch(error) {
                    error({
                        code: GeolocationPositionError.POSITION_UNAVAILABLE,
                        message: "Failed to parse location object from MiniAppBridge: " + error
                    })
                }
```

What do you think?

js-miniapp-bridge/src/bridge.js
```diff
@@ -114,6 +114,31 @@ MiniAppBridge.requestPermission = function(permissionType) {
...
+        return MiniAppBridge.exec(
+            "getCurrentPosition",
+            {locationOptions: options},
+            function(value) {
+                try {
+                    var parsedData = JSON.parse(value)
+                    success(parsedData)
+                } catch(error) {
+                    error({
+                        code: GeolocationPositionError.POSITION_UNAVAILABLE,
```
minh-rakuten, line 132: I'm not sure about this. Can we call `GeolocationPositionError` directly like this here?
@ts-leojoseph-r @corycaywood

ts-leojoseph-r: @minh-rakuten . Yes we can -> https://w3c.github.io/geolocation-api/#position_error_interface

PR #31, opened 2020-08-04
refactor: TypeScript bridge

minh-rakuten:
# Description
Convert js bridge into TypeScript.
CI for js bridge bundle. Mobile SDK will use JS SDK as submodule with `sparse-checkout`.

## Links
[MINI-759](https://jira.rakuten-it.com/jira/browse/MINI-759)

# Checklist


js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -0,0 +1,120 @@
...
+export interface Callback {
+  id: string;
+
+  onSuccess: (value: string) => void;
+  onError: (error: string) => void;
+}
+
+export class MiniAppBridge {}
+
+export interface MiniAppBridge {
```
corycaywood, line 14: I think we don't need to define the `MiniAppBridge` interface and class separately. You can define the typing at the same time as defining the class, and also the method return values can be inferred by Typescript.

The only problem is that the `exec` function needs to be implemented by the platform specific code. So maybe you could create a separate `PlatformExecutor` which is provided when constructing the `MiniAppBridge`?

Something like this in `common-bridge.ts`:
```
export interface PlatformExecutor {
  exec(
    action: string,
    param: any,
    onSuccess: (value: string) => void,
    onError: (error: string) => void
  ): void;
}

export class MiniAppBridge {
  executor: PlatformExecutor;

  constructor(executor: PlatformExecutor) {
    this.executor = executor;
  }

  requestPermission(permissionType: string) {
    return new Promise<string>((resolve, reject) => {
      return this.executor.exec(
        // ...etc
      );
    });
  }

  // ...etc
}
```

And then implement it in `android-bridge.ts`:
```
class AndroidExcecutor implements PlatformExecutor {
  exec(action, param, onSuccess, onError) {
      // ...
    );
  }
}

(window as any).MiniAppBridge = new MiniAppBridge(new AndroidExcecutor());
```

What do you think?

js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -0,0 +1,120 @@
...
+  getUniqueId(): Promise<string>;
+
+  /**
+   * Associating requestPermission function to MiniAppBridge object
+   * @param {String} permissionType Type of permission that is requested. For eg., location
+   */
+  requestPermission(permissionType: string): void;
+}
+
+MiniAppBridge.prototype.execSuccessCallback = (messageId, value) => {
```
corycaywood, line 61: Adding to the `prototype` works, but it seems we lose some of the benefits of Typescript. See my above comment about defining these as class methods and then instantiating `new MiniAppBridge()`

.circleci/config.yml
```diff
@@ -79,6 +79,27 @@ jobs: # a collection of steps
       - run:
           name: Lint
           command: yarn bridge lint
+      - run:
```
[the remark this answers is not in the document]
minh-rakuten: Thank you!

js-miniapp-bridge/src/common-bridge.ts
minh-rakuten: I think good to go that way.

PR #39, opened 2020-08-21
bridge: add custom permission function in common bridge

khairul-alam-licon:
# Description
Add `requestCustomPermissions` function in `common-bridge.ts` to ask for the custom permissions from miniapps to Android, iOS or Native.

# Jira
[MINI-959](https://jira.rakuten-it.com/jira/browse/MINI-959)

# Checklist

### DIFF
--- a/js-miniapp-bridge/src/common-bridge.ts
+++ b/js-miniapp-bridge/src/common-bridge.ts
@@ -102,7 +102,7 @@ export class MiniAppBridge {
 
   /**
    * Associating requestCustomPermissions function to MiniAppBridge object
-   * @param string[] permissionTypes, Type of custom permissions that is requested
+   * @param [string, string] permissionTypes, Type of custom permissions that is requested
    * using an Array including the parameters eg. name, description.
    *
    * Miniapps should use "rakuten.miniapp.user." prefix for a valid name of the custom permissions
--- a/js-miniapp-bridge/src/common-bridge.ts
+++ b/js-miniapp-bridge/src/common-bridge.ts
@@ -103,6 +103,7 @@ export class MiniAppBridge {
   /**
    * Associating requestCustomPermissions function to MiniAppBridge object
    * @param {string[]} permissionTypes, Array of type of custom permissions that is requested
+   * For eg., user name, profile photo, contact list.
    */
   requestCustomPermissions(permissionTypes: string[]) {
     return new Promise<string>((resolve, reject) => {
--- a/js-miniapp-bridge/src/common-bridge.ts
+++ b/js-miniapp-bridge/src/common-bridge.ts
@@ -87,13 +87,29 @@ export class MiniAppBridge {
 
   /**
    * Associating requestPermission function to MiniAppBridge object
-   * @param {String} permissionType Type of permission that is requested. For eg., location
+   * @param {string} permissionType Type of permission that is requested. For eg., location
    */
   requestPermission(permissionType: string) {
     return new Promise<string>((resolve, reject) => {
       return this.executor.exec(
         'requestPermission',
         { permission: permissionType },
+        success => resolve(success),
+        error => reject(error)
+      );
+    });
+  }
+
+  /**
+   * Associating requestCustomPermissions function to MiniAppBridge object
+   * @param {string[]} permissionTypes, Array includes type of custom permissions that is requested. 
+   * For eg., user name, profile photo, contact list
+   */
+  requestCustomPermissions(permissionTypes: string[]) {
+    return new Promise<string>((resolve, reject) => {
+      return this.executor.exec(
+        'requestCustomPermissions',
+        { permissions: permissionTypes },
         success => resolve(success),
         error => reject(error)
       );
--- a/js-miniapp-bridge/src/common-bridge.ts
+++ b/js-miniapp-bridge/src/common-bridge.ts
@@ -102,8 +102,7 @@ export class MiniAppBridge {
 
   /**
    * Associating requestCustomPermissions function to MiniAppBridge object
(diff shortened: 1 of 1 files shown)


js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -99,6 +99,21 @@ export class MiniAppBridge {
       );
     });
   }
+
+  /**
+   * Associating requestCustomPermissions function to MiniAppBridge object
+   * @param {string[]} permissionTypes, Array of type of custom permissions that is requested
```
ts-leojoseph-r, line 105: Array.<string> would be better. Also please try to add an example that could be much more helpful.

js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -99,6 +99,21 @@ export class MiniAppBridge {
...
+   */
+  requestCustomPermissions(permissionTypes: string[]) {
```
ts-leojoseph-r, line 107: I guess we don't need to specify the datatype that is passed.

ts-leojoseph-r, line 107: Also, I guess we need to have one more parameter for a description (which will be used by the host app while showing the custom permissions request)
`"The developer must specify a reason for each permission request."`
Please check [MINI-952](https://jira.rakuten-it.com/jira/browse/MINI-952)

minh-rakuten: In TypeScript, we need to specify the data type.
TS is defined as type object-oriented compile language which prevents the javascript any type style.

ts-leojoseph-r: Ah, I almost forgot that we changed to Typescript. My bad, sorry. @khairul-alam-licon please ignore.

khairul-alam-licon: I think, we can work with the "reason" parameter in a separate PR? Is it possible?

js-miniapp-bridge/src/common-bridge.ts
khairul-alam-licon: I have added an example, let me to know if it's okay, or we can do something better.

js-miniapp-bridge/src/common-bridge.ts
corycaywood: The `description` will passed within the string array, so I don't think anything needs to be done here.

js-miniapp-bridge/src/common-bridge.ts
corycaywood: Yeah, I think this needs an example of the string array that should be passed.

corycaywood (approved):
Looks good 👍  But I think we need an example of how the params are passed to `requestCustomPermissoions`.

js-miniapp-bridge/src/common-bridge.ts
ts-leojoseph-r: That's right @corycaywood , we actually changed the request  string object after the comment here. So I don't think any change is required here.

js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -99,6 +99,35 @@ export class MiniAppBridge {
...
+   *  {"name":"rakuten.miniapp.user.PROFILE_PHOTO", "description": "Reason to request for the custom permission"},
+   *  {"name":"rakuten.miniapp.user.CONTACT_LIST", "description": "Reason to request for the custom permission"}
+   * ]
+   *
+   * For eg., Miniapps can pass the array of invalid custom permissions as following
+   * [
+   *  {"name":"any unknown name", "description": "Reason to request for the custom permission"}
+   * ]
+   */
+  requestCustomPermissions(permissionTypes: [string, string]) {
```
corycaywood, line 121: I think that `[string, string]` defines an Array that contains two string values, i.e. `['first string', 'second string']`. But this is supposed to be an array that contains objects, right? So I think it should be something like this:
```suggestion
  requestCustomPermissions(permissionTypes: Array<{name: string, description: string}>) {
```

js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -99,6 +99,35 @@ export class MiniAppBridge {
...
+   *  {"name":"rakuten.miniapp.user.USER_NAME", "description": "Reason to request for the custom permission"},
+   *  {"name":"rakuten.miniapp.user.PROFILE_PHOTO", "description": "Reason to request for the custom permission"},
+   *  {"name":"rakuten.miniapp.user.CONTACT_LIST", "description": "Reason to request for the custom permission"}
+   * ]
+   *
+   * For eg., Miniapps can pass the array of invalid custom permissions as following
+   * [
+   *  {"name":"any unknown name", "description": "Reason to request for the custom permission"}
+   * ]
+   */
```
minh-rakuten, line 120: I think we can remove those lines.

js-miniapp-bridge/src/common-bridge.ts
minh-rakuten: How about creating an object for this? `Array<CustomPermission>`?

corycaywood: Yeah, making a `CustomPermission` type is even better.

corycaywood (approved):
👍

js-miniapp-bridge/src/common-bridge.ts
khairul-alam-licon: Thanks, I have updated.

minh-rakuten (approved):
👍 👍

PR #44, opened 2020-09-02
refactor: brightcove video & remove mobile restriction

minh-rakuten:
# Description
- Update Media section with brightcove video which is similar to what MAB is using. The previous media was not good.

- Remove mobile restriction when fetching unique id. Android & iOS also run on tablet, watch, etc not only mobile.

## Links
MINI-1615
MINI-1616
MINI-1619

# Checklist


.circleci/config.yml
```diff
@@ -134,7 +134,6 @@ jobs: # a collection of steps
           command: |
             yarn sample prettify
             yarn sample lint
-            yarn sample flow
```
corycaywood, line 137: That's one solution 😆 

I think you can add the custom flow typing at `js-miniapp-sample/flow-typed/npm/@brightclove/react-player-loader_vx.x.x.js`, and then you just have to define this content in the file:

```
declare module '@brightcove/react-player-loader' {
    declare module.exports: any;
}
```

Actually, maybe we can convert this project to TypeScript sometime in the future? It would make sense because the other projects in the repo are using TypeScript.

js-miniapp-sample/src/pages/media.js
```diff
@@ -1,14 +1,18 @@
...
 const Media = () => {
   return (
-    <video width="100%" controls>
-      <source
-        src="https://multiplatform-f.akamaihd.net/i/multi/will/bunny/big_buck_bunny_,640x360_400,640x360_700,640x360_1000,950x540_1500,.f4v.csmil/master.m3u8"
-        type="application/x-mpegURL"
-      ></source>
-    </video>
+    <ReactPlayerLoader
+      accountId="1752604059001"
```
corycaywood, line 12: Where does this come from? Is it just a test video provided by brightclove?

corycaywood (approved):
Looks good 👍  But probably it's better to re-enable flow so we can keep that check. See below.

js-miniapp-sample/src/pages/media.js
minh-rakuten: Yes, I got this info from the official example of brightcove.
https://player.support.brightcove.com/coding-topics/react-player-loader.html

nitish-a-jain (dismissed):
I think the JS SDK Sample App should continue to use default tags to showcase their support. And not integrate a paid solution like Brightcove.
Note: It is a proprietory player which one of the customers wants to use.

Climbatize (approved):
I share concerns about using a third party video from a commercial company. But this is just a demo app that provides examples of what can be done to our customers, and our actual customers needing this feature are using BrightCove, so lets validate this PR so it can help them advance and let's update it later with something more conventional if needed

Abhijit-majumder (approved):
looks good to me!

PR #45, opened 2020-09-07
ci: Publish JS bridge to branches instead of master

corycaywood:
# Description
Previously we were using git sparse-checkout in order to import a single bridge.js file to the Android project. However, on iOS it's not possible to use sparse-checkout with Cocoapods, so this solution won't work on iOS. Instead, we will publish the built bridge.js file to platform-specific branches which contain only a single file: js-bridge-android and js-bridge-ios

# Checklist


.circleci/config.yml
```diff
@@ -1,4 +1,28 @@
...
+      platform:
+        description: Platform to export - android or ios
+        type: string
+    steps: 
+      - run: 
+          name: Publish <<parameters.platform>>
+          command: |
+            set -e
+            git checkout js-bridge-<<parameters.platform>>
+            cp js-miniapp-bridge/build/src/<<parameters.platform>>/bridge.js ./
```
minh-rakuten, line 15: This one `js-miniapp-bridge/build/` is put in `persist_to_workspace` so is it still existed even we do `git checkout branch`?

corycaywood: Yeah, it should be because the `build` folder is part of `.gitignore`, so it should still be there when switching branches. I tested it on my fork, and it seemed to work: 
https://app.circleci.com/pipelines/github/corycaywood/js-miniapp/44/workflows/5e203dec-179c-4840-8412-4fc0ff981ad9/jobs/92

minh-rakuten (approved):
👍 👍

khairul-alam-licon (approved):
👍 👍

PR #46, opened 2020-09-09
feature: share info

minh-rakuten:
# Description
- SDK:
Share info feature. For now, only attribute content.

- Bridge:
Share info feature
Refactor: import CustomPermissionType from `js-miniapp-sdk`.

The shared info is an object so we will add more properties for it when there are more requirements on data sharing.

# Checklist


js-miniapp-sdk/src/miniapp.ts
```diff
@@ -10,6 +12,13 @@ interface MiniAppFeatures {
 
   /** @returns The Promise of permission result of mini app from injected side. */
   requestLocationPermission(): Promise<string>;
+
+  /**
+   * @param info The shared data must match the property in [ShareInfoType].
+   * @returns The Promise of load ad response result from injected side.
+   * Promise is rejected if failed to load.
+   */
+  shareInfo(info: ShareInfoType): Promise<null | Error>;
```
corycaywood, line 21: On another PR I had suggested including `Error` in the type, but looks like this isn't working correctly. My editor isn't able to detect the return type correctly. So I think maybe this should be just `Promise<string>`?

corycaywood (approved):
Looks good 👍  But seems there are some conflicts which need to be resolved now.

js-miniapp-sdk/src/miniapp.ts
minh-rakuten: I have changed into `Promise<undefined>`. What do you think?

We don't return anything onSuccess because the miniapp just emits the shared info object.
The error string can be handled in onError so I guess we don't need `Error` in type.

js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -195,6 +192,22 @@ export class MiniAppBridge {
...
+   * Associating shareInfo function to MiniAppBridge object.
+   * This function does not return anything back on success.
+   * @param {info} The shared info object.
+   */
+  shareInfo(info: ShareInfoType) {
+    return new Promise<undefined>((resolve, reject) => {
+      return this.executor.exec(
+        'shareInfo',
+        info,
+        success => resolve(undefined),
```
ts-leojoseph-r, line 206: Do you think we can return something like a `SUCCESS` would be better?  We can return SUCCESS once we pass the string to the Host app delegate/callback. What do you think ?

js-miniapp-sdk/src/miniapp.ts
```diff
@@ -10,6 +12,13 @@ interface MiniAppFeatures {
 
   /** @returns The Promise of permission result of mini app from injected side. */
   requestLocationPermission(): Promise<string>;
+
+  /**
+   * @param info The shared data must match the property in [ShareInfoType].
+   * @returns The Promise of load ad response result from injected side.
```
ts-leojoseph-r, line 18: I think `returns` description need to be updated

js-miniapp-sdk/test/miniapp.spec.ts
```diff
@@ -173,3 +174,28 @@ describe('loadRewardedAd', () => {
...
+  });
+  it('should retrive error response from the MiniAppBridge once there is errors', () => {
+    const sharedInfo = {
+      content: 'test content',
+    };
+
+    const error: Error = {
+      message: 'Unknown error occured',
+      name: 'Bridge error',
+    };
```
ts-leojoseph-r, line 196: @minh-rakuten Is this the standard error format for all bridge communication ?? We are trying to follow a generic error format in iOS, something like below,
```
{
   "error":{
      "title":"ERROR_TITLE",
      "description":"DESCRIPTION OF THE ERROR"
   }
}
```

ts-leojoseph-r (commented):
Looks good to me, just a few comments to discuss

js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -195,6 +192,22 @@ export class MiniAppBridge {
...
+  /**
+   * Associating shareInfo function to MiniAppBridge object.
+   * This function does not return anything back on success.
+   * @param {info} The shared info object.
+   */
+  shareInfo(info: ShareInfoType) {
+    return new Promise<undefined>((resolve, reject) => {
+      return this.executor.exec(
+        'shareInfo',
+        info,
```
ts-leojoseph-r, line 205: Can we change this `info` to `shareInfo` or anything you feel that is a little more descriptive? Because I guess iOS works a little differently than android in this case. We need to add this variable in **[RequestParameters](https://github.com/rakutentech/ios-miniapp/blob/master/MiniApp/Classes/Models/MiniAppJavaScriptMessageInfo.swift#L10)**

js-miniapp-sdk/src/miniapp.ts
minh-rakuten: Updating to `Promise<string>`.

js-miniapp-sdk/test/miniapp.spec.ts
minh-rakuten: Actually, the error format both android & bridge using is just the only string.

js-miniapp-bridge/src/common-bridge.ts
minh-rakuten: Yeah let me change it.

js-miniapp-sdk/test/miniapp.spec.ts
corycaywood: I think we should use the standard [JavaScript error format](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) - it uses `message` and `name` keys.

The [MDN docs for Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject) make it sound like using an instance of `Error` is the best practice.

minh-rakuten: Sounds like we need to change on android, ios and bridge.
I will change the bridge code for `Error` in another PR.

ts-leojoseph-r (approved):
Looks good to me :+1:

PR #51, opened 2020-09-15
fix: (bridge) Parse json response on request permissions

corycaywood:
# Description
The iOS/Android side returns a string so this must be first parsed before it can be used. 
Also did some refactoring of the bridge tests to reduce duplication.

## Links
MINI-1586

# Checklist


minh-rakuten (approved):
👍 👍

ts-leojoseph-r (approved):
Looks good to me :+1:

PR #52, opened 2020-09-16
refactor: Move types to bridge package

corycaywood:
# Description
This allows us to define these types at a lower level and have a one-way dependency on the bridge types from 'js-miniapp-sdk'. This also lets us use the 'MiniAppBridge' type in the 'js-miniapp-sdk' code to ensure that the SDK is conforming to what the bridge code expects

I also had to change the sdk/bridge typescript and build settings a little bit so that the types from `js-miniapp-bridge` will be available in the published npm module without needing a dependency on `js-miniapp-bridge` itself.

# Checklist


minh-rakuten (commented):
In my perspective view, the one way dependency here is contradicted 🤔 
`js-bridge` toward `js-sdk` is better to me. The bridge is between `mobile-sdk` & `js-sdk` so it should not hold models.

corycaywood:
> In my perspective view, the one way dependency here is contradicted 🤔
> `js-bridge` toward `js-sdk` is better to me. The bridge is between `mobile-sdk` & `js-sdk` so it should not hold models.

The biggest advantage of doing this is that we get type safety when calling the Bridge code from the JS SDK code. Right now, we're casting `MiniAppBridge` as `any` which means that we have no guarantee that the Bridge and JS SDK code are compatible with each other. If we move the types into the Bridge, then we can import `MiniAppBridge` into the JS SDK side and are guaranteed that we're using the `MiniAppBridge` correctly.

But the problem is that we can't have the types separated from the bridge if we want to do that. If we did, then we would end up having a two way dependency: the JS SDK depending on Bridge so that we can ensure we're using the bridge code correctly - but also the Bridge depending on the JS SDK so that we can use the types in the bridge code. This is bad practice and also leads to compilation errors.

So I think this reveals a flaw with having the types in the JS SDK side. They should instead be in the bridge because this is what directly interacts with the native side, and the JS SDK is just a layer on top of the bridge. If there are types which are used only by the JS SDK, then those can be in the JS SDK side rather than the bridge. 
@minh-rakuten @Climbatize @ts-leojoseph-r  What do you think?

ts-leojoseph-r:
> > In my perspective view, the one way dependency here is contradicted 🤔
> > `js-bridge` toward `js-sdk` is better to me. The bridge is between `mobile-sdk` & `js-sdk` so it should not hold models.
> 
> The biggest advantage of doing this is that we get type safety when calling the Bridge code from the JS SDK code. Right now, we're casting `MiniAppBridge` as `any` which means that we have no guarantee that the Bridge and JS SDK code are compatible with each other. If we move the types into the Bridge, then we can import `MiniAppBridge` into the JS SDK side and are guaranteed that we're using the `MiniAppBridge` correctly.
> 
> But the problem is that we can't have the types separated from the bridge if we want to do that. If we did, then we would end up having a two way dependency: the JS SDK depending on Bridge so that we can ensure we're using the bridge code correctly - but also the Bridge depending on the JS SDK so that we can use the types in the bridge code. This is bad practice and also leads to compilation errors.
> 
> So I think this reveals a flaw with having the types in the JS SDK side. They should instead be in the bridge because this is what directly interacts with the native side, and the JS SDK is just a layer on top of the bridge. If there are types which are used only by the JS SDK, then those can be in the JS SDK side rather than the bridge.
> @minh-rakuten @Climbatize @ts-leojoseph-r What do you think?

Gosh, sorry I forgot to review and add my comments on this PR.
Honestly I feel that it is better to have Types closer to the Native SDK i.e in the bridge code which not only enables closer integrity but also it avoids unnecessary type checks. Also, the JS SDK code was meant for encapsulation not the other way around, so I guess this approach is good. We can provide more wrappers or utils from the JS SDK though.

ts-leojoseph-r (approved):
Looks good to me :+1:

Abhijit-majumder (approved):
Looks good!! 👍

minh-rakuten (approved):
👍

PR #53, opened 2020-09-28
refactor: admob

minh-rakuten:
# Description
SDK:
- Update the response type of ads to string, Reward.
- Device OS detection. The value could be `Android`, `iOS` or `Unknown`.

Bridge:
- Update the response type of ads to string, Reward.
- Platform detection. The value could be `Android` or `iOS`.

# Checklist

### DIFF
--- a/js-miniapp-bridge/src/android/android-bridge.ts
+++ b/js-miniapp-bridge/src/android/android-bridge.ts
@@ -20,6 +20,10 @@ class AndroidExcecutor implements PlatformExecutor {
       JSON.stringify({ action, param, id: callback.id })
     );
   }
+
+  getPlatform(): string {
+    return 'Android';
+  }
 }
 
 (window as any).MiniAppBridge = new MiniAppBridge(new AndroidExcecutor());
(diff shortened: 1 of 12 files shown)


js-miniapp-bridge/src/android/android-bridge.ts
```diff
@@ -20,6 +20,10 @@ class AndroidExcecutor implements PlatformExecutor {
       JSON.stringify({ action, param, id: callback.id })
     );
   }
+
+  getPlatform(): string {
+    return 'Android';
```
corycaywood, line 25: Can we make an `enum` for this? So we could use something like `Platform.ANDROID` here.

corycaywood (approved):
Looks good 👍  Just left a suggestion

Abhijit-majumder (approved):
Looks Good 👍 
Like Cory mentioned, please put possible platforms to enum
Thanks

ts-leojoseph-r (approved):
Looks good to me :+1:

PR #57, opened 2020-10-03
Retrieve Username and Profile photo

ts-leojoseph-r:
# Description
* Bridging script changes to retrieve UserName and ProfilePhoto
* JS-SDK script changes to retrieve UserName and ProfilePhoto
* js-sdk-sample app changes to retrieve and display the user details

# Checklist

### DIFF
--- a/js-miniapp-bridge/src/common-bridge.ts
+++ b/js-miniapp-bridge/src/common-bridge.ts
@@ -227,6 +227,40 @@ export class MiniAppBridge {
       );
     });
   }
+
+  /**
+   * Associating getUsername function to MiniAppBridge object.
+   * This function returns username from the user profile
+   * (provided the rakuten.miniapp.user.USER_NAME is allowed by the user)
+   * It returns error info if user had denied the custom permission
+   */
+  getUserName() {
+    return new Promise<string>((resolve, reject) => {
+      return this.executor.exec(
+        'getUserName',
+        null,
+        success => resolve(success),
+        error => reject(error)
+      );
+    });
+  }
+
+  /**
+   * Associating getProfilePhoto function to MiniAppBridge object.
+   * This function returns username from the user profile
+   * (provided the rakuten.miniapp.user.PROFILE_PHOTO is allowed by the user)
+   * It returns error info if user had denied the custom permission
+   */
+  getProfilePhoto() {
+    return new Promise<string>((resolve, reject) => {
+      return this.executor.exec(
+        'getProfilePhoto',
+        null,
+        success => resolve(success),
+        error => reject(error)
+      );
+    });
+  }
 }
 
 /**
--- a/js-miniapp-bridge/src/common-bridge.ts
+++ b/js-miniapp-bridge/src/common-bridge.ts
@@ -229,9 +229,9 @@ export class MiniAppBridge {
   }
 
   /**
-   * Associating getUsername function to MiniAppBridge object.
+   * Associating getUserName function to MiniAppBridge object.
    * This function returns username from the user profile
-   * (provided the rakuten.miniapp.user.USER_NAME is allowed by the user)
+   * (provided the rakuten.miniapp.user.USER_NAME custom permission is allowed by the user)
    * It returns error info if user had denied the custom permission
    */
   getUserName() {
--- a/js-miniapp-bridge/src/common-bridge.ts
+++ b/js-miniapp-bridge/src/common-bridge.ts
@@ -239,7 +239,7 @@ export class MiniAppBridge {
       return this.executor.exec(
         'getUserName',
         null,
-        success => resolve(success),
+        userName => resolve(userName),
         error => reject(error)
       );
     });
@@ -256,7 +256,7 @@ export class MiniAppBridge {
       return this.executor.exec(
         'getProfilePhoto',
         null,
-        success => resolve(success),
+        profilePhoto => resolve(profilePhoto),
         error => reject(error)
       );
     });
--- a/js-miniapp-sample/src/services/user/actions.js
+++ b/js-miniapp-sample/src/services/user/actions.js
@@ -15,7 +15,7 @@ type ContactListSuccessAction = { type: String, contacts: string[] };
 
 const requestUserName = (): Function => {
   return (dispatch) => {
-    MiniApp.getUserName()
+    return MiniApp.getUserName()
     .then((userName) => {
       dispatch({
         type: REQUEST_USER_NAME_SUCCESS,
@@ -32,7 +32,7 @@ const requestUserName = (): Function => {
 
 const requestProfilePhoto = (): Function => {
   return (dispatch) => {
-    MiniApp.getProfilePhoto()
+    return MiniApp.getProfilePhoto()
     .then((url) => {
       dispatch({
         type: REQUEST_PROFILE_PHOTO_SUCCESS,
--- a/js-miniapp-sample/src/services/user/actions.js
+++ b/js-miniapp-sample/src/services/user/actions.js
@@ -1,3 +1,5 @@
+import MiniApp from 'js-miniapp-sdk';
+
 import {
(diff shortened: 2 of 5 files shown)


js-miniapp-sdk/src/miniapp.ts
```diff
@@ -35,6 +35,16 @@ interface MiniAppFeatures {
    * @returns The Promise of share info action state from injected side.
    */
   shareInfo(info: ShareInfoType): Promise<string>;
+
+  /**
+   * @returns Username saved in the host app user profile
+   */
+  getUserName(): Promise<string>;
```
corycaywood, line 42: We're starting to have a lot of methods in `MiniApp` (and will have a lot more to come). So I was thinking, maybe we should start name-spacing them or somehow providing a logical separation between related functionality. This would allow us to separate them in the source code and also provide a more clear interface to the user (hopefully).

We've kind of started to do that with the `Ad` and `Platform` interfaces, but I think we could take this even further by moving each interface into its own separate implementation which is then set as a property of `MiniApp`. So basically I mean something like this:

```
interface UserInfoProvider {
  getUserName(): Promise<string>;
}

/** @internal */
class UserInfo implements UserInfoProvider {
  getUserName(): Promise<string> {
    return this.bridge.getUserName();
  }
}

export class MiniApp {
  user: UserInfoProvider = new UserInfo();
}
```

Or another solution is to completely separate `UserInfoProvider` into its own class which must be imported and used on its own, but that doesn't really fit with what we've done so far of having everything under `MiniApp`.

What do you think?

corycaywood (commented):
It's looking good 👍  I've left a few comments.

js-miniapp-sample/src/services/user/actions.js
```diff
@@ -13,41 +15,35 @@ type ContactListSuccessAction = { type: String, contacts: string[] };
...
-          userName,
-        });
-
-        return Promise.resolve(userName);
-      })
-      .catch((_) => {
-        dispatch({
-          type: REQUEST_USER_NAME_FAILURE,
-        });
+    MiniApp.getUserName()
```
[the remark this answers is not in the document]
ts-leojoseph-r: Done

js-miniapp-sample/src/services/user/actions.js
```diff
@@ -13,41 +15,35 @@ type ContactListSuccessAction = { type: String, contacts: string[] };
...
-          url,
-        });
-
-        return Promise.resolve(url);
-      })
-      .catch((_) => {
-        dispatch({
-          type: REQUEST_PROFILE_PHOTO_FAILURE,
-        });
+    MiniApp.getProfilePhoto()
```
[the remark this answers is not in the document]
ts-leojoseph-r: Done

js-miniapp-sdk/src/miniapp.ts
ts-leojoseph-r: Honestly I wanted to separate `UserInfoProvider` into its own class, import and work. Not just `UserInfoProvider` but also other interfaces.  I thought I will do refactor in another ticket

js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -227,6 +227,40 @@ export class MiniAppBridge {
...
+   * This function returns username from the user profile
+   * (provided the rakuten.miniapp.user.USER_NAME is allowed by the user)
+   * It returns error info if user had denied the custom permission
+   */
+  getUserName() {
+    return new Promise<string>((resolve, reject) => {
+      return this.executor.exec(
+        'getUserName',
+        null,
+        success => resolve(success),
```
minh-rakuten, line 242: It will return username directly so I think we can replace `success` with `username`. The same for `getProfilePhoto()`.

minh-rakuten (approved):
Nice!
Just a minor thing before the approve.

js-miniapp-bridge/src/common-bridge.ts
ts-leojoseph-r: Done

js-miniapp-sdk/README.md
```diff
@@ -162,6 +162,34 @@ miniApp.shareInfo(info)
...
+
+MiniApp.getUserName()
+    .then(userName => {
+		console.log(userName);
+	}).catch(error => {
+		console.error(error);
+	});
+```
+
+NOTE: getProfilePhoto() -  Returns the Base64 Encoded image string from the Host app.
```
khairul-alam-licon, line 180: Is it possible to include any convention or example as a guideline for the host app e.g.

```
data:image/png;base64,XXX
```

js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -227,6 +227,40 @@ export class MiniAppBridge {
       );
     });
   }
+
+  /**
+   * Associating getUsername function to MiniAppBridge object.
+   * This function returns username from the user profile
+   * (provided the rakuten.miniapp.user.USER_NAME is allowed by the user)
```
khairul-alam-licon, line 234: I think, we can say-

```
@returns User name from the user profile if the required permission has been granted
e.g. "rakuten.miniapp.user.USER_NAME" 
```

khairul-alam-licon (approved):
Looks good, also I left few ideas.

js-miniapp-sdk/README.md
minh-rakuten: oops now I noticed this. It's possible for the hostapp to return the net url for the profile photo so I think it can be 
```suggestion
NOTE: getProfilePhoto() -  Returns the image uri from the Host app.
```
The miniapp can load both http url or base64 string (with correct format).

@ts-leojoseph-r @khairul-alam-licon in our android & ios sample app, we return the base64 string so we need to pass it with the form `data:image/png;base64, XXX`.

ts-leojoseph-r: That's right @minh-rakuten . We are alreaddy returning as base 64 encoded string from iOS to the miniapp. Thanks for the suggestion, updated the comments.

js-miniapp-bridge/src/common-bridge.ts
ts-leojoseph-r: Done

minh-rakuten (approved):

js-miniapp-sdk/src/miniapp.ts
```diff
@@ -81,9 +81,42 @@ interface Platform {
...
+
+  getProfilePhoto(): Promise<string> {
+    return this.bridge.getProfilePhoto();
+  }
+}
+
 /* tslint:disable:no-any */
 export class MiniApp implements MiniAppFeatures, Ad, Platform {
   private bridge: MiniAppBridge = (window as any).MiniAppBridge;
+  private user: UserInfoProvider = new UserInfo(this.bridge);
```
corycaywood, line 119: I was thinking that this property would be public, so the user data features would be used by calling `MiniApp.user.getUserName()`, and then you don't need to add a third `getUserName` function to this class. But honestly, I'm not sure what's the best interface design to use.

What do you think?

ts-leojoseph-r: I think its a good idea too. I couldn't imagine the number of interfaces added to MiniAppBridge. Sooner the better we break down like this. I think it is good to have it for `Ad` as well before MiniApp starts using them.

corycaywood (approved):
Looks good 👍 

As we discussed, let's think about how we can re-design the way that public interfaces are accessed for `UserInfoProvider`, `Ads`, and `Platform` and address it in another PR. Whether it's by adding properties for each in `MiniApp`, or adding separate public classes for each, or having instantiate functions like `MiniApp.user()` (see [firebase example](https://firebase.google.com/docs/reference/js/firebase.analytics)). I'm not sure what's the best solution, but we need a way to both keep our own code organized and to provide a logical separation in the public interface for the SDK users.

Abhijit-majumder (approved):
👍

PR #58, opened 2020-10-05
fetch-api-wrapper

Rajesh-Sundaram:
# Description
add fetch-api-wrapper to js-sdk to make http request at host-application.

# Checklist


minh-rakuten (commented):
It's so good but there is the issue of data type in cross platform.
I believe android & ios do not support `RequestInit` and `Response` of typescript so we need to define the data model in bridge.

The format of request and response you can take a look at `Mini App Web Requests to External APIs` internal document page. Please feel free to ping me or Cory if you can't find it.

js-miniapp-bridge/src/types/fetch.ts
```diff
@@ -0,0 +1,4 @@
+export type MiniAppRequestInit = Pick<
+  RequestInit,
```
corycaywood, line 2: I agree with @minh-rakuten about this - We should explicitly specify our own type rather than relying on part of the browser's/typescript's implementation. We are using the browser's `fetch` as inspiration, but there will end up being some differences.

js-miniapp-bridge/src/types/fetch.ts
```diff
@@ -0,0 +1,4 @@
...
+  'body' | 'headers' | 'method'
```
corycaywood, line 3: We're starting out with support only for `GET` requests, so there's no need for `body` currently, right?

js-miniapp-bridge/src/types/fetch.ts
```diff
@@ -0,0 +1,4 @@
+export type MiniAppRequestInit = Pick<
```
corycaywood, line 1: How about the name `FetchRequest`? I don't think we need to add the `MiniApp` prefix to our types because the user will be able rename them when importing, if needed: i.e. `import {WebRequest as MiniAppRequest} from 'js-miniapp-sdk'`

js-miniapp-sdk/src/miniapp.ts
```diff
@@ -35,6 +36,15 @@ interface MiniAppFeatures {
...
   shareInfo(info: ShareInfoType): Promise<string>;
+
+  /**
+   * makes the http request at Host application layer.
+   *
+   * @param url requestUrl
+   * @param options requestOptions
+   * @returns success or failure http response.
+   */
+  fetch(url: string, options: MiniAppRequestInit): Promise<Response>;
```
corycaywood, line 47: I think we should explicitly define the `Response` type too. Maybe it can be named `FetchResponse`?

Also, shouldn't `options` be optional? I wonder if we should use the same style as the browser's `fetch` where it can be called as any of the following:
- `fetch("https://example.com")`
- `fetch("https://example.com", {headers: headers})`
- `fetch({url: "https://example.com", headers: headers})`

What do you think?

corycaywood (commented):
It's looking good 👍  I've left a few comments.

js-miniapp-sdk/src/miniapp.ts
Abhijit-majumder: I agree with @corycaywood here that we should explicitly define the response interface.

js-miniapp-bridge/src/types/fetch.ts
Rajesh-Sundaram: If we don't have any security implications on providing POST method support, can we have the POST method implementation on the native-side with a simple string (raw-data, JSON) as a request body for the initial-phase?

corycaywood: Hmm, I'm not sure - there aren't really any security concerns, but this will add more complexity on the native side, which we wanted to avoid for this initial implementation.

On the JS side, it's not very difficult to add support for POST requests, so maybe we can leave it out for now and then it will still be easy to add in JS when we decide to support it on the native side.

js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -261,6 +273,22 @@ export class MiniAppBridge {
...
+   *
+   * @param request request informations
+   */
+  fetch(request: FetchRequest) {
+    return new Promise<NativeFetchResponse>((resolve, reject) => {
+      return this.executor.exec(
+        'fetch',
+        request,
+        success => resolve(JSON.parse(success) as NativeFetchResponse),
+        error => reject(JSON.parse(error) as NativeFetchResponse)
```
corycaywood, line 288: I think `error` should pass a string and not a `Response`. This is because the Promise will only be rejected if there was some issue with sending the request or receiving a response (i.e. the user set invalid params, tried to connect to a URL which isn't in the "Allow List", device is not connected to network, etc.) So in that case, there will be no `Response` object to return to the user.

If the server returns an error response (i.e. 500), then the Promise will be resolved with the error response - this is why the `Response` model has an `ok` property to check if the response is successful.

corycaywood (commented):
Looking good! 👍  I left some comments below.

js-miniapp-bridge/src/types/fetch.ts
```diff
@@ -0,0 +1,111 @@
...
+  text(): Promise<string>;
+}
+
+export class FetchResponseImpl implements Response {
+  private encoding = 'utf-8';
+  headers: Record<string, string>;
+  ok: boolean;
+  status: number;
+  statusText: string;
+  type: ResponseType;
```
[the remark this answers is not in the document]
Rajesh-Sundaram: 👍

js-miniapp-sdk/src/miniapp.ts
```diff
@@ -165,4 +183,18 @@ export class MiniApp implements MiniAppFeatures, Ad, Platform {
...
+
+  async fetch(
+    input: FetchRequest | string,
+    init?: FetchRequestInit
+  ): Promise<FetchResponse> {
+    let request;
+    if (typeof input === 'string') {
+      request = { url: input as string, ...init };
+    } else {
+      request = { ...input, ...init };
```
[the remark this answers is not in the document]
Rajesh-Sundaram: FetchRequestInit doesn't have `url` property. init option can be used to override the headers & method.
if we ignore the second arg, can we throw an error/warning message when the user passes both object args?

js-miniapp-sdk/src/miniapp.ts
```diff
@@ -165,4 +183,18 @@ export class MiniApp implements MiniAppFeatures, Ad, Platform {
     } catch (e) {}
     return platform;
   }
+
+  async fetch(
```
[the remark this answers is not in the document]
Rajesh-Sundaram: in progress 👍

js-miniapp-bridge/src/common-bridge.ts
Rajesh-Sundaram: 👍  makes sense. Thanks

js-miniapp-bridge/src/types/fetch.ts
```diff
@@ -0,0 +1,111 @@
...
+   * Decode response as json
+   */
+  json(): Promise<object>;
+  /**
+   * Decode response as text
+   */
+  text(): Promise<string>;
+}
+
+export class FetchResponseImpl implements Response {
```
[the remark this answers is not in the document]
Rajesh-Sundaram: would it be okay if we keep this implementations of bridge-types in js-SDK workspace?

js-miniapp-sdk/src/miniapp.ts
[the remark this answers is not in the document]
corycaywood: Hmm, maybe we should just simplify things and only support `fetch(input: string, opts?: FetchRequestInit)`? 

Do you think there are any benefits to supporting `fetch(input: FetchRequest | string, opts?: FetchRequestInit)`? The only thing I can think of is similarity to the browser's `fetch` function, but our `fetch` will already be different anyways.

js-miniapp-bridge/src/types/fetch.ts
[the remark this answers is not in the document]
corycaywood: I don't think that the `sdk` workspace is the right place for this. The `js-miniapp-sdk` npm module is mainly intended as a wrapper for the bridge functionality, so we try not to put much in there beyond what we have to.

Also, keeping this in the `bridge` workspace couples it with the native side because the bridge JS gets bundled with the native SDK. So if we make changes to the native side then we can simultaneously makes changes to the bridge JS. This is important in case we need to make changes in the bridge which would otherwise be breaking for mini apps that included an old version of `js-miniapp-sdk`.

js-miniapp-sdk/src/miniapp.ts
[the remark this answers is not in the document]
Rajesh-Sundaram: I absolutely agree with this simplification 👍

js-miniapp-bridge/src/types/fetch.ts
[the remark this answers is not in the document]
Rajesh-Sundaram: 👍 Got it. Thanks for clarifying

js-miniapp-bridge/src/fetch/response.ts
```diff
@@ -0,0 +1,50 @@
+import { NativeFetchResponse, FetchResponse } from '../types/fetch';
+
+export class DecodedFetchResponse implements FetchResponse {
+  private encoding = 'utf-8';
+  headers: Record<string, string>;
+  ok: boolean;
+  status: number;
+  statusText: string;
+  url: string;
+  body: Uint8Array;
```
minh-rakuten, line 10: Why do we have to declare again when `Response` and `ResponeBody` have them already?

minh-rakuten (commented):
I just only have questions due to my lack of knowledge.

corycaywood (commented):
Looks good - just a few issues with the public interface and generated docs.

js-miniapp-bridge/src/fetch/response.ts
Rajesh-Sundaram: since DecodedFetchResponse extends the Response type, it's required to implement all of its parent properties in its class.

js-miniapp-sdk/src/miniapp.ts
```diff
@@ -165,4 +181,10 @@ export class MiniApp implements MiniAppFeatures, Ad, Platform {
     } catch (e) {}
     return platform;
   }
+
+  async fetch(url: string, opts?: FetchRequestInit): Promise<FetchResponse> {
```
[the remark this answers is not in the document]
Rajesh-Sundaram: 👍  async/await not required as this bridge returns promise
updated with below
`fetch(url: string, opts?: FetchRequestInit): Promise<FetchResponse> {
    return this.bridge
      .fetch(new InternalFetchRequest(url, opts))
      .then(res => new DecodedFetchResponse(res));
  }`

minh-rakuten (approved):
👍 
There could be the update for this when native sdk implements this feature but we will figure it out later.

corycaywood (commented):
Looks good 👍 

But I think we shouldn't merge this yet because I'm not sure when we want to release this. Or maybe you can create a new feat/fetch branch for merging this PR there?

Abhijit-majumder (approved):
Looks Good 👍

Rajesh-Sundaram:
> Looks good 👍
> 
> But I think we shouldn't merge this yet because I'm not sure when we want to release this. Or maybe you can create a new feat/fetch branch for merging this PR there?

@corycaywood 👍 Done. Can I merge?

corycaywood (approved):
👍

PR #62, opened 2020-10-12
Sample app updates for QA testability

corycaywood:
# Description
- sample: (ads) Add input fields and 'load' buttons
  - This is needed so all test cases can be covered in QA
- sample: (uri) Add external webview page with tel and _blank links
  - This HTML file is accessed via the raw github url.

# Checklist

### DIFF
--- a/js-miniapp-sample/external-webview/index.html
+++ b/js-miniapp-sample/external-webview/index.html
@@ -1,0 +1,52 @@
+<!DOCTYPE html>
+<!-- This page is used on the "URI Schemes" page as an example for opening the external webview -->
+<!-- It is accessed directly via the raw link to this file on Github -->
+<html>
+    <head>
+        <title>External Webview</title>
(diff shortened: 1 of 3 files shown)


khairul-alam-licon (approved):
👍

minh-rakuten (commented):
Looks good!

Only one thing about the error display in ad app. It's better to have the two error displays for two types of ad, also we can show the error message from sdk instead of showing the default error message.

corycaywood:
> Looks good!
> 
> Only one thing about the error display in ad app. It's better to have the two error displays for two types of ad, also we can show the error message from sdk instead of showing the default error message.

Thanks, this is a good idea. I have updated with these changes. Also, the JSX portion was getting pretty long (and had some duplicate code) so I refactored a little to have separate render methods.

minh-rakuten (approved):
👍 👍

PR #68, opened 2020-10-29
feature: retrieve access token

minh-rakuten:
# Description
- Get access token from hostapp as `TokenData` object
- Update the authentication mobile section to demo this feature: Display token and expiration date.

Note: As a discussion, the user permission for access token will be implemented when first time permission is done.

## Links
MINI-1865

# Checklist

### DIFF
--- a/js-miniapp-bridge/src/common-bridge.ts
+++ b/js-miniapp-bridge/src/common-bridge.ts
@@ -13,7 +13,7 @@ import {
 } from './types/custom-permissions';
 import { ShareInfoType } from './types/share-info';
 import { ScreenOrientation } from './types/screen';
-import { AccessTokenData } from './types/token-data';
+import { NativeTokenData, AccessTokenData } from './types/token-data';
 
 /** @internal */
 const mabMessageQueue: Callback[] = [];
@@ -272,7 +272,10 @@ export class MiniAppBridge {
       return this.executor.exec(
         'getAccessToken',
         null,
-        tokenData => resolve(JSON.parse(tokenData) as AccessTokenData),
+        tokenData => {
+          const nativeTokenData = JSON.parse(tokenData) as NativeTokenData;
+          resolve(new AccessTokenData(nativeTokenData));
+        },
         error => reject(error)
       );
     });
--- a/js-miniapp-bridge/src/common-bridge.ts
+++ b/js-miniapp-bridge/src/common-bridge.ts
@@ -13,6 +13,7 @@ import {
 } from './types/custom-permissions';
 import { ShareInfoType } from './types/share-info';
 import { ScreenOrientation } from './types/screen';
+import { TokenData } from './types/token-data';
 
 /** @internal */
 const mabMessageQueue: Callback[] = [];
@@ -258,6 +259,20 @@ export class MiniAppBridge {
         'getProfilePhoto',
         null,
         profilePhoto => resolve(profilePhoto),
+        error => reject(error)
+      );
+    });
+  }
+
+  /**
+   * Get access token from native hostapp.
+   */
+  getAccessToken() {
+    return new Promise<TokenData>((resolve, reject) => {
+      return this.executor.exec(
+        'getAccessToken',
+        null,
+        tokenData => resolve(JSON.parse(tokenData) as TokenData),
         error => reject(error)
       );
     });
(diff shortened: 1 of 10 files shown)


js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -263,6 +264,20 @@ export class MiniAppBridge {
...
 
+  /**
+   * Get access token from native hostapp.
+   */
+  getAccessToken() {
+    return new Promise<TokenData>((resolve, reject) => {
+      return this.executor.exec(
+        'getAccessToken',
+        null,
+        tokenData => resolve(JSON.parse(tokenData) as TokenData),
```
ts-leojoseph-r, line 275: We send the JSON string from the respective Android/iOS SDK's, do we still need to `JSON.parse` ?

https://github.com/rakutentech/ios-miniapp/pull/161/files#diff-fd17266a8049bc8366e4f1fba97d7a5bc0568eb62b4d3a2bf10122daefbe90deR211

minh-rakuten: Yeah we need it.
The native sdk sends the json string and the js sdk parse that json string into object. The object here is `TokenData`.

js-miniapp-bridge/src/types/token-data.ts
```diff
@@ -0,0 +1,5 @@
+/** Token data type. */
+export interface TokenData {
```
corycaywood, line 2: Maybe the name `AccessTokenData` would be better to match the function name `getAccessToken`? 

Also, we may have other types of tokens in the future, such as an ID Token (Open ID Connect), so it's good to be specific on the token type.

js-miniapp-bridge/src/types/token-data.ts
```diff
@@ -0,0 +1,5 @@
...
+  token: string;
+  validUntil: number;
```
corycaywood, line 4: Maybe we can return this as a `Date` object? It would be more convenient for the users and would ensure that the bridge is receiving the same date format on both platforms. What do you think?

corycaywood (commented):
Looking good, just a couple of suggestions.

js-miniapp-bridge/src/types/token-data.ts
minh-rakuten: I think the `long` type is more generic for cross-platforms. @ts-leojoseph-r  What do you think?

js-miniapp-bridge/src/types/token-data.ts
minh-rakuten: I have updated to `AccessTokenData` 👍

js-miniapp-bridge/src/types/token-data.ts
corycaywood: Sorry, I think I commented this in the wrong place - I agree it's best to return a `long` over the JS bridge. But I think we could convert it to a `Date` before returning the object to the user. But I guess that might require two different types to be defined - one for the object received over the bridge and one for the object returned to the user.

However, if we return the date as a `number`, then we need to add information in the docs about the format of the date, i.e. number of milliseconds since the Unix Epoch.

js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -263,6 +264,20 @@ export class MiniAppBridge {
...
 
+  /**
+   * Get access token from native hostapp.
+   */
+  getAccessToken() {
+    return new Promise<AccessTokenData>((resolve, reject) => {
+      return this.executor.exec(
+        'getAccessToken',
+        null,
+        tokenData => resolve(JSON.parse(tokenData) as AccessTokenData),
```
khairul-alam-licon, line 275: I am not sure, if only `resolve(JSON.parse(tokenData))` here is enough?

khairul-alam-licon (commented):
Looks good so far.

js-miniapp-bridge/src/types/token-data.ts
ts-leojoseph-r: I guess `long` type is okay but at the same time it is good to convert the same to `Date`

js-miniapp-bridge/src/common-bridge.ts
minh-rakuten: I have updated. Define the type here help on clearing the context 😄.

corycaywood (approved):
Looks good! 👍

khairul-alam-licon (approved):
👍 👍

PR #75, opened 2020-11-27
feat: add location permission to custom permissions management

khairul-alam-licon:
# Description
- Add a custom permission type for `rakuten.miniapp.device.location`.
- We need to consider how this permission fits into the flow of the native device permission for LOCATION. This PR aims to request for the device location using `requestCustomPermissions` first before it will ask by `requestPermission`. It will add the solution inside sdk as default implementation instead of sample app.

## Links
- MINI-1956

# Checklist

### DIFF
--- a/js-miniapp-bridge/src/types/custom-permissions.ts
+++ b/js-miniapp-bridge/src/types/custom-permissions.ts
@@ -2,6 +2,7 @@ export enum CustomPermissionName {
   USER_NAME = 'rakuten.miniapp.user.USER_NAME',
   PROFILE_PHOTO = 'rakuten.miniapp.user.PROFILE_PHOTO',
   CONTACT_LIST = 'rakuten.miniapp.user.CONTACT_LIST',
+  LOCATION = 'rakuten.miniapp.device.LOCATION'
 }
 
 export enum CustomPermissionStatus {
--- a/js-miniapp-bridge/src/types/custom-permissions.ts
+++ b/js-miniapp-bridge/src/types/custom-permissions.ts
@@ -2,7 +2,7 @@ export enum CustomPermissionName {
   USER_NAME = 'rakuten.miniapp.user.USER_NAME',
   PROFILE_PHOTO = 'rakuten.miniapp.user.PROFILE_PHOTO',
   CONTACT_LIST = 'rakuten.miniapp.user.CONTACT_LIST',
-  LOCATION = 'rakuten.miniapp.device.LOCATION'
+  LOCATION = 'rakuten.miniapp.device.LOCATION',
 }
 
 export enum CustomPermissionStatus {
--- a/js-miniapp-sdk/CHANGELOG.md
+++ b/js-miniapp-sdk/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### In Progress (2020-XX-XX)
 - **Feature:** Added `CustomPermissionName.LOCATION`.
-- **Change:** Updated `requestLocationPermission()`` to `requestLocationPermission(permissionDescription?: string)`. From now `requestLocationPermission` will request both custom and device permission respectively.
+- **Change:** Updated `requestLocationPermission()` to `requestLocationPermission(permissionDescription?: string)`. From now `requestLocationPermission` will request both custom and device permission respectively.
 
 ### 1.5.0 (2020-11-13)
 
--- a/js-miniapp-sdk/CHANGELOG.md
+++ b/js-miniapp-sdk/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### In Progress (2020-XX-XX)
 - **Feature:** Added `CustomPermissionName.LOCATION`.
-- **Change:** Updated `requestLocationPermission()`` to `requestLocationPermission(permissionDescription: string)`. From now `requestLocationPermission` will request both custom and device permission respectively.
+- **Change:** Updated `requestLocationPermission()`` to `requestLocationPermission(permissionDescription?: string)`. From now `requestLocationPermission` will request both custom and device permission respectively.
 
 ### 1.5.0 (2020-11-13)
 
--- a/js-miniapp-sdk/CHANGELOG.md
+++ b/js-miniapp-sdk/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### In Progress (2020-XX-XX)
 - **Feature:** Added `CustomPermissionName.LOCATION`.
-- **Change:** Updated `requestLocationPermission()`` to `requestLocationPermission(description?: string)`. From now `requestLocationPermission` will request both custom and device permission respectively.
+- **Change:** Updated `requestLocationPermission()`` to `requestLocationPermission(permissionDescription: string)`. From now `requestLocationPermission` will request both custom and device permission respectively.
 
 ### 1.5.0 (2020-11-13)
 
--- a/js-miniapp-sdk/CHANGELOG.md
+++ b/js-miniapp-sdk/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## CHANGELOG
 
 ### In Progress (2020-XX-XX)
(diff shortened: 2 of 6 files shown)


js-miniapp-sdk/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## CHANGELOG
 
+### In Progress (2020-XX-XX)
+- **Feature:** Added `rakuten.miniapp.device.LOCATION` custom permission.
```
corycaywood, line 4: `rakuten.miniapp.device.LOCATION` is more like an implementation detail, so I think it should be like this instead:
```suggestion
- **Feature:** Added `CustomPermissionName.LOCATION` custom permission.
```

js-miniapp-sdk/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## CHANGELOG
 
+### In Progress (2020-XX-XX)
```
corycaywood, line 3: Maybe you should also mention the update to `requestLocationPermission` since it will now request both the device and custom permission?

js-miniapp-bridge/src/types/custom-permissions.ts
```diff
@@ -2,6 +2,7 @@ export enum CustomPermissionName {
   USER_NAME = 'rakuten.miniapp.user.USER_NAME',
   PROFILE_PHOTO = 'rakuten.miniapp.user.PROFILE_PHOTO',
   CONTACT_LIST = 'rakuten.miniapp.user.CONTACT_LIST',
+  LOCATION = 'rakuten.miniapp.device.LOCATION',
```
corycaywood, line 5: I think this also needs to be added to the user guide for JS SDK.

js-miniapp-sdk/src/miniapp.ts
```diff
@@ -146,7 +148,22 @@ export class MiniApp implements MiniAppFeatures, Ad, Platform {
   }
 
   requestLocationPermission(): Promise<string> {
-    return this.requestPermission(DevicePermission.LOCATION);
+    const locationPermission = [
+      {
+        name: CustomPermissionName.LOCATION,
+        description: 'We would like to display the location of your device.',
```
corycaywood, line 154: I think we should allow the developer to optionally pass the description, so maybe we could change this method to `requestLocationPermission(description?: string)`.

Also, I don't think we should assume what the purpose of requesting the location is, so could we make the description blank by default?

corycaywood (commented):
Looking good 👍  I left a few comments below.

js-miniapp-sdk/src/miniapp.ts
khairul-alam-licon: Thanks Cory, I'll update this part as well.

js-miniapp-sdk/CHANGELOG.md
khairul-alam-licon: Yeah, I have updated this part.

js-miniapp-sdk/src/miniapp.ts
```diff
@@ -145,8 +147,23 @@ export class MiniApp implements MiniAppFeatures, Ad, Platform {
     return this.bridge.getUniqueId();
   }
 
-  requestLocationPermission(): Promise<string> {
-    return this.requestPermission(DevicePermission.LOCATION);
+  requestLocationPermission(permissionDescription?: string): Promise<string> {
```
corycaywood, line 150: This also needs to be updated in the `MiniAppFeature` interface above (https://github.com/rakutentech/js-miniapp/pull/75/files#diff-07ef2fcb929b435ee80df92f95af63de439887918ea0710a246b4fc14fb3220dL20).

Also, this will pass `undefined` to the native SDK if the user doesn't define `permissionDescription`. Will the native side handle this properly? You could instead use a default value like `permissionDesciption: string = ""`.

khairul-alam-licon: Ok, according to typescript, I have updated this part.

js-miniapp-sdk/test/miniapp.spec.ts
```diff
@@ -43,12 +43,21 @@ describe('getUniqueId', () => {
...
+    ],
+  });
+
   it('should delegate to requestPermission function when request any permission', () => {
     const spy = sinon.spy(miniApp, 'requestPermission' as any);
 
     miniApp.requestLocationPermission();
 
-    return expect(spy.callCount).to.equal(1);
+    return expect(spy.callCount).to.equal(0);
```
minh-rakuten, line 60: You need to make this test case work properly 😆

corycaywood: You can use the expectation with a promise like this:
```
return miniApp.requestLocationPermission()
  .then(() -> expect(...))
```

js-miniapp-sdk/src/miniapp.ts
```diff
@@ -17,7 +19,7 @@ interface MiniAppFeatures {
   getUniqueId(): Promise<string>;
 
   /** @returns The Promise of permission result of mini app from injected side. */
-  requestLocationPermission(): Promise<string>;
+  requestLocationPermission(permissionDescription: string): Promise<string>;
```
corycaywood, line 22: I think this still needs the optional: `permissionDescription?: string`. Or else this would be a breaking change (for Typescript users).

js-miniapp-bridge/src/types/custom-permissions.ts
khairul-alam-licon: Yeah, I have updated `js-miniapp-sdk/README.md ` to address this suggestion.

js-miniapp-sdk/test/miniapp.spec.ts
khairul-alam-licon: Thanks, I have fixed this test.

js-miniapp-sdk/src/miniapp.ts
khairul-alam-licon: Sorry, I forgot about to make it optional. I have updated this part with `permissionDescription?: string`.

js-miniapp-sdk/README.md
```diff
@@ -70,8 +70,9 @@ Note that there are two types of permissions:
 
 | Permission | Type | Method | Description |
 | --- | --- | --- | --- |
-| LOCATION | Device | `requestLocationPermission()` | The current location of the device.<br>Data position can be accessible via geolocation request (`navigator.geolocation`). |
-| User Data | Custom | `requestCustomPermission()` | Custom permissions for User Data. Currently supported types are `CustomPermissionName.USER_NAME`, `CustomPermissionName.PROFILE_PHOTO`, and `CustomPermissionName.CONTACT_LIST`. These permisions should be requested before you attempt to access the User's data. |
+| Location | Device | `requestLocationPermission()` | The current location permission of the device.<br>Data position can be accessible via geolocation request (`navigator.geolocation`). |
+| Location | Custom | `requestCustomPermissions()` | Custom permission for device location. Currently supported type is `CustomPermissionName.LOCATION`. |
```
minh-rakuten, line 74: I think you can remove this line and add `Custom permission for device location. Currently supported type is CustomPermissionName.LOCATION` to the section above. Also, we need to update the `Device permissions` content [here](https://github.com/rakutentech/js-miniapp/blob/3d1279368c2fe735033f783ad5e5af34aa944762/js-miniapp-sdk/README.md#3-request-permissions) to explain that device permission is also custom permission now, it is required custom permission to be processed first.

corycaywood (approved):
Looks good 👍

js-miniapp-sdk/README.md
khairul-alam-licon: I have updated this part. 

> Also, we need to update the Device permissions content here to explain that device permission is also custom permission now.

Ah, I think, not all device permissions require custom permissions except location at the moment.

minh-rakuten (approved):
👍

PR #76, opened 2020-11-27
ci: Deploy sample app to gh-pages branch on master merge

corycaywood:
# Description
This is so we can test the JS Sample App as a website using the new SDK feature.

I'm a little bit sad about using the `gh-pages` branch for multiple purposes (hosting both docs and the sample app) - but I think this is the most simple and straightforward way of deploying. I guess it's just a disadvantage of using a mono-repo.

# Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,10 +23,54 @@ commands:
               echo "Nothing to commit"
             fi
 
-jobs: # a collection of steps
-  build-sdk: # runs not using Workflows must have a `build` job as entry point
-    working_directory: ~/mini-js-sdk # directory where steps will run
-    docker: # run the steps with Docker
+jobs: 
+  build-bridge:
+    working_directory: ~/mini-js-bridge
+    docker:
+      - image: circleci/node:10.16.3
+    steps:
+      - checkout
+      - restore_cache:
+          name: Restore Yarn Package Cache
+          key: yarn-packages-{{ checksum "yarn.lock" }}
+      - run:
+          name: Install dependencies w/ Yarn
+          command: yarn install --frozen-lockfile
+      - save_cache:
+          name: Save Yarn Package Cache
+          key: yarn-packages-{{ checksum "yarn.lock" }}
+          paths:
+            - ~/.cache/yarn
+      - run:
+          name: Test
+          command: yarn bridge test
+      - run:
+          name: Lint
+          command: yarn bridge lint
+      - run:
+          name: Compile
+          command: yarn bridge compile
+      - persist_to_workspace:
+          root: ~/mini-js-bridge
+          paths:
+            - js-miniapp-bridge/build/
+
+  export-bridge:
+    working_directory: ~/mini-js-bridge
+    docker:
+      - image: circleci/node:10.16.3
+    steps:
+      - checkout
+      - attach_workspace:
+          at: ./
+      - export-js-bridge:
+          platform: ios
+      - export-js-bridge:
+          platform: android
+
+  build-sdk:
+    working_directory: ~/mini-js-sdk
+    docker:
       - image: circleci/node:10.16.3
     steps:
       - checkout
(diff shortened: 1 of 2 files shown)


nitish-a-jain (commented):
Hi @corycaywood, Just a few quick questions:

- What is the level of urgency for this feature? 
- Is it temporary?
- What are the alternatives?

We could support to host the sample mini-app on a proper domain as well. Few of our subscriptions are round the corner as well.

corycaywood:
> Hi @corycaywood, Just a few quick questions:
> 
> * What is the level of urgency for this feature?
> * Is it temporary?
> * What are the alternatives?
> 
> We could support to host the sample mini-app on a proper domain as well. Few of our subscriptions are round the corner as well.

@nitish-a-jain To answer your questions:

* We would like to test this as soon as possible so the feature can be released in the next 1-2 weeks
* This isn't intended to be temporary
* Alternatives would be something like Firebase or Azure

Do you see any disadvantages to this approach?

The only disadvantage that I saw is that we will have the Github pages site being used for hosting both docs and the sample app. However, the sample app will be in a separate folder, so it shouldn't affect the docs in any way or vice versa.

Also using Github pages has these advantages over Firebase/Azure:
* Don't need to create/manage new credentials or add credentials to the CI job
* Free with no resource usage limit
* Simple to manage, and everything for JS SDK/Sample is completely contained in github

What do you think?

nitish-a-jain:
> > Hi @corycaywood, Just a few quick questions:
> > 
> > * What is the level of urgency for this feature?
> > * Is it temporary?
> > * What are the alternatives?
> > 
> > We could support to host the sample mini-app on a proper domain as well. Few of our subscriptions are round the corner as well.
> 
> @nitish-a-jain To answer your questions:
> 
> * We would like to test this as soon as possible so the feature can be released in the next 1-2 weeks
> * This isn't intended to be temporary
> * Alternatives would be something like Firebase or Azure
> 
> Do you see any disadvantages to this approach?
> 
> The only disadvantage that I saw is that we will have the Github pages site being used for hosting both docs and the sample app. However, the sample app will be in a separate folder, so it shouldn't affect the docs in any way or vice versa.
> 
> Also using Github pages has these advantages over Firebase/Azure:
> 
> * Don't need to create/manage new credentials or add credentials to the CI job
> * Free with no resource usage limit
> * Simple to manage, and everything for JS SDK/Sample is completely contained in github
> 
> What do you think?

Thanks for sharing a bit on the background. Though I am still not sure about the _basic need_; as hosting a mini-app goes beyond the scope of being a mini-app itself. And probably the website will look weird with two **totally** different contexts on it.

If it is to do with internal testing, probably, it could be approached differently. A **temporary** attempt to **repurpose** for **urgency** could probably make sense. Let's discuss this, as scheduled.

corycaywood:
> Thanks for sharing a bit on the background. Though I am still not sure about the _basic need_; as hosting a mini-app goes beyond the scope of being a mini-app itself. And probably the website will look weird with two **totally** different contexts on it.
> 
> If it is to do with internal testing, probably, it could be approached differently. A **temporary** attempt to **repurpose** for **urgency** could probably make sense. Let's discuss this, as scheduled.

@nitish-a-jain Sorry, I forgot to mention the context here. In the demo app, we've added the functionality to display a mini app from a URL, including the full functionality of the mini app (see this PR: https://github.com/rakutentech/android-miniapp/pulls). This is so that mini app developers can preview their mini app within the demo app just by running a local server.

So the purpose of hosting the JS Sample App as a website is so that we can easily QA this new "Load from URL" feature of the demo app.

.circleci/config.yml
```diff
@@ -157,6 +156,33 @@ jobs: # a collection of steps
...
+              git add sample
+              git config user.name "CI Publisher"
+              git config user.email "dev-opensource@mail.rakuten.com"
+              git commit -m "sample: Deploy JS Sample App [ci skip]"
+              git push origin gh-pages
+            else
+              echo "Nothing to commit"
+            fi
 
   publish-sdk:
```
minh-rakuten, line 187: Now our build & publish flow is sdk -> bridge -> sample. Since there are some object models being exported from js bridge so I think the flow should be bridge -> sdk -> sample.

.circleci/config.yml
```diff
@@ -132,6 +129,8 @@ jobs: # a collection of steps
     docker:
       - image: circleci/node:12.16.2
       # TODO get the sdk build from saved workspace
+    environment:
+      PUBLIC_URL: .
```
minh-rakuten, line 133: I don't know what it is ^^. Could you tell me what is it for?

minh-rakuten (commented):
Thanks to this PR, I think we have a problem with the order of bridge and sdk build. The `bridge` holds object models so it should be built first.

.circleci/config.yml
corycaywood: This environment variable used by create-react-app and it appends it to each link for JS and CSS files. See here: https://create-react-app.dev/docs/advanced-configuration/

The problem is that create-react-app assumes that the App is hosted on the root of the domain, but ours will not be on the root (it will be under `https://rakutentech.github.io/js-miniapp/sample`) . So I set this to `.` so that links to assets will be like `./script.js`.

.circleci/config.yml
corycaywood: I think that the jobs will run in parallel, so the order in the config file just determines what order they show in the UI on CircleCI. But I noticed that the jobs weren't really grouped together, so I rearranged them a to follow bridge -> sdk -> sample.

minh-rakuten (approved):
👍 👍

ts-leojoseph-r (approved):
Looks good to me :+1:

PR #79, opened 2020-12-02
doc: reformat the link reference

minh-rakuten:
# Description
Fix the broken link of ShareInfoType and update other links with published page format.

## Links
MINI-2200

# Checklist


js-miniapp-sdk/README.md
```diff
@@ -147,7 +147,7 @@ miniApp.loadRewardedAd(adUnitID)
 
 It is possible for the mini app user to share data with another App by showing the native content sharing chooser.
 
-The data format must match the [ShareInfoType](https://github.com/rakutentech/js-miniapp/blob/master/js-miniapp-sdk/src/types/ShareInfoType.ts).
+The data format must match the [ShareInfoType](../js-miniapp-bridge/src/types/share-info.ts).
```
corycaywood, line 150: This should point to the generated docs on the github pages site. For example, the link needs to work when they click the link from here: https://rakutentech.github.io/js-miniapp/

So I think the link be something like `api/interfaces/shareinfotype.html` (which will end up pointing to https://rakutentech.github.io/js-miniapp/docs/1.5/api/interfaces/shareinfotype.html)

minh-rakuten: Yeah but everytime the version jumps up, it does not refer to latest version, i.e the next version is 1.6.

corycaywood: Yeah, that's kind of the problem - you can have a relative link that either works from the published user guide (`api/interfaces/shareinfotype.html`) or works from the file viewed via the github repo (`../js-miniapp-bridge/src/types/share-info.ts`), but not a relative link which works in both places. 

But the published user guide is supposed to be the main entry point for developers who are integrating the SDK, so I think it makes more sense to have it work there, i.e. `api/interfaces/shareinfotype.html`.

Or I guess we could point directly to the full github.com href like `https://www.github.com/js-miniapp-bridge/src/types/share-info.ts`?

minh-rakuten: I have updated this relative link format `api/interfaces/shareinfotype.html` but it does not seem work. Do I miss something 🤔 ?

corycaywood: Unfortunately, I think it will only work after it's been published to https://rakutentech.github.io/js-miniapp 😣 . I think you can also change the linke to point to the generated `.md` file (api/interfaces/shareinfotype.md) and then check the docs locally with `yarn sdk docs`. The `.md` link should be converted to `.html` by Github after it's been published.

It's also possible to have TypeDoc automatically generate the API links in the readme by using `[[ShareInfoType]]`. But it doesn't work correctly with our current setup - so maybe we can look into this in the future.

corycaywood: Also, you can test the Github generated docs locally by copying the generated docs to the `gh-pages` branch and running `bundle exec jekyll serve` (similar as explained in the Android docs: https://github.com/rakutentech/android-miniapp#how-to-generate-kdocs-sdk-documentation-locally)

corycaywood (approved):
Looks good 👍 

As I mentioned, you could also point the links to the `.md` file to make it easier to test locally if you want.

khairul-alam-licon (approved):
👍

PR #80, opened 2020-12-07
docs: Fix name appearing in generated typedocs

corycaywood:
# Description
The name was showing up in typedocs as 'js-miniapp-bridge' and also it was showing the version number for js-miniapp-bridge instead of the version for js-miniapp-sdk. This is possibly a bug in typedocs where it's pulling this from an external module rather than the main module. The only fix I could find is to manually set the 'name' property and remove the version.

Also moved the typedoc options into a typedoc.json file

# Checklist

### DIFF
--- a/js-miniapp-sdk/scripts/docs.sh
+++ b/js-miniapp-sdk/scripts/docs.sh
@@ -17,7 +17,7 @@ version: "$PACKAGE_VERSION"
 cat <<< "$contents" > "$FILE_VERSION_MD"
 
 # Generate docs
-npx typedoc --includeVersion --out $DIR_DOCS/api src --plugin typedoc-plugin-markdown --stripInternal --readme none --mode file
+npx typedoc --out $DIR_DOCS/api src --options typedoc.json
 
 # Move userguide to docs and create correct folder structure
 mv $DIR_DOCS/api/README.md $DIR_DOCS/api/index.md
--- a/js-miniapp-bridge/package.json
+++ b/js-miniapp-bridge/package.json
@@ -21,8 +21,6 @@
     "source-map-support": "^0.5.16",
     "ts-node": "^8.8.1",
     "tslint": "^6.1.0",
-    "typedoc": "^0.17.3",
-    "typedoc-plugin-markdown": "^2.2.17",
     "typescript": "^3.8.3"
   },
   "scripts": {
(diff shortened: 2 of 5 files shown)


minh-rakuten (commented):
@corycaywood 
Screen Shot 2020-12-08 at 16 52 21
Do you mean this issue?

corycaywood:
> @corycaywood
> Screen Shot 2020-12-08 at 16 52 21
> Do you mean this issue?

Yeah, that's the one. I think it's because we have some source files from js-miniapp-bridge included in the generated docs. But it's strange because js-miniapp-bridge an external module and not the main module, so I'm not sure why it uses that as the name 🤔

js-miniapp-sdk/scripts/docs.sh
```diff
@@ -17,7 +17,7 @@ version: "$PACKAGE_VERSION"
 cat <<< "$contents" > "$FILE_VERSION_MD"
 
 # Generate docs
-npx typedoc --includeVersion --out $DIR_DOCS/api src --plugin typedoc-plugin-markdown --stripInternal --readme none --mode file
+npx typedoc --out $DIR_DOCS/api src --options typedoc.json
```
minh-rakuten, line 20: when I generated the doc locally, I still see the version of js-miniapp-bridge 1.1.0

minh-rakuten (commented):
I think because our data model objects are from bridge repository so the document of those objects are labeled with js-miniapp-bridge.

js-miniapp-sdk/scripts/docs.sh
corycaywood: Oh strange, did you use `yarn sdk docs`? Also, did you delete the `publishableDocs` folder before running it? I think sometimes it doesn't generate correctly if you don't delete the folder first.

I checked on the CircleCI build and it looks like the name is correct in the Artifacts:
https://app.circleci.com/pipelines/github/corycaywood/js-miniapp/123/workflows/a3d0937f-6160-4c28-8dd1-26552fa46960/jobs/307/artifacts

minh-rakuten: you're right, _delete the publishableDocs folder before running_ works.

minh-rakuten (approved):
👍

khairul-alam-licon (approved):
👍 👍

PR #81, opened 2020-12-08
refactor: change bridge call

minh-rakuten:
# Description
Stop using local bridge variable to avoid early injection.

## Links
MINI-2148

# Checklist

### DIFF
--- a/js-miniapp-sdk/src/miniapp.ts
+++ b/js-miniapp-sdk/src/miniapp.ts
@@ -11,6 +11,8 @@ import {
   AccessTokenData,
 } from '../../js-miniapp-bridge/src';
 
+/* tslint:disable:no-any */
+
 /**
  * A module layer for webapps and mobile native interaction.
  */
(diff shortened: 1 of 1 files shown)


js-miniapp-sdk/src/miniapp.ts
```diff
@@ -118,36 +120,28 @@ export interface UserInfoProvider {
...
-  }
-
   getUserName(): Promise<string> {
-    return this.bridge.getUserName();
+    return (window as any).MiniAppBridge.getUserName();
   }
 
   getProfilePhoto(): Promise<string> {
-    return this.bridge.getProfilePhoto();
+    return (window as any).MiniAppBridge.getProfilePhoto();
```
corycaywood, line 128: Using `any` will make us lose the typing benefits of TypeScript, so we don't know whether we're accidentally trying to use a function on `MiniAppBridge` which doesn't exist. 

Could you instead create a function which dynamically returns the typed `MiniAppBridge`? For example:

```
function getBridge() {
  return (window as any).MiniAppBridge as MiniAppBridge;
}
```

minh-rakuten: Yes, I have updated.

corycaywood (approved):
LGTM 👍

khairul-alam-licon (approved):
👍 👍

ts-leojoseph-r (approved):
Looks good to me :+1:

PR #83, opened 2020-12-11
External webview: allow parameters to be passed and received

corycaywood:
# Description
This was done by updating `external-webview/index.html` to allow receiving parameters and also sending a "callback URL" to this page (the URL of the mini app).

On the external webview page, the user can also set parameters which will be returned to the JS SDK Sample App, and the sample app will display these parameters on the landing page.

# Checklist

### DIFF
--- a/js-miniapp-sample/src/pages/uri-schemes.js
+++ b/js-miniapp-sample/src/pages/uri-schemes.js
@@ -1,16 +1,21 @@
 // @flow
-import React from 'react';
+import React, { useState } from 'react';
 
 import {
   Button,
   CardContent,
   CardActions,
+  TextField,
   makeStyles,
 } from '@material-ui/core';
 
 import GreyCard from '../components/GreyCard';
 
 const useStyles = makeStyles((theme) => ({
+  actions: {
+    justifyContent: 'center',
+    paddingBottom: 16,
+  },
   card: {
     height: 'auto',
   },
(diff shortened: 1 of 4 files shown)


js-miniapp-sample/src/pages/uri-schemes.js
```diff
@@ -24,14 +29,41 @@ const useStyles = makeStyles((theme) => ({
...
+
+  function onOpenExternalWebview() {
+    if (params && !validateParams(params)) {
+      window.alert(
+        'Invalid params. Please input params in the format ?param1=value1&param2=value2'
+      );
+      return;
+    }
+
+    const callbackUrl = `${window.location.protocol}//${window.location.host}/index.html`;
```
minh-rakuten, line 59: It is more ideal to redirect to `URI Schemes` page
but this redirect to landing page is okay if we don't want to spend much effort on sample app 😄

minh-rakuten (approved):
👍 👍

khairul-alam-licon (approved):
Looks good 👍

js-miniapp-sample/src/pages/uri-schemes.js
corycaywood: Yeah, that's true - this is much simpler to do 😆 

Also, we can re-use this for showing params that are passed to the mini app when created.

PR #100, opened 2021-02-01
sample: send message to single contact

minh-rakuten:
# Description
Sample app for send message to single contact based on Message to chatbot screen.

## Links
MINI-1650

# Checklist


khairul-alam-licon (approved):
Looks good 👍

corycaywood (approved):
Looks good! 👍 

I'm just thinking maybe we should rename everything with the word `ChatBot`. It seems like it could be confusing because it's not actually a chat bot. Could it be replaced with `Message` and `MessageType`?

PR #110, opened 2021-03-18
Display Name & Email in Contact list

ts-leojoseph-r:
# Description
* Display Name & Email in the Contact list screen.
* For old Contacts which has only ID - Name & Email will not be displayed.

## Links
MINI-2914

Screen Shot 2021-03-19 at 0 06 31

# Checklist

### DIFF
--- a/js-miniapp-sample/src/pages/user-details.js
+++ b/js-miniapp-sample/src/pages/user-details.js
@@ -318,13 +318,15 @@ function UserDetails(props: UserDetailsProps) {
                   primary={contact.id}
                   secondary={
                     <React.Fragment>
-                      <Typography>{contact.name !== "" ? (
-                        <span>{"Name: " + contact.name}</span>
-                      ) : null}
+                      <Typography>
+                        {contact.name !== '' ? (
+                          <span>{'Name: ' + contact.name}</span>
+                        ) : null}
                       </Typography>
-                      <Typography>{contact.email !== "" ? (
-                        <span>{"Email: " + contact.email}</span>
-                      ) : null}
+                      <Typography>
+                        {contact.email !== '' ? (
+                          <span>{'Email: ' + contact.email}</span>
+                        ) : null}
                       </Typography>
                     </React.Fragment>
                   }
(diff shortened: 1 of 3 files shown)


minh-rakuten (commented):
Looks good.

You need to update [Contact](https://github.com/rakutentech/js-miniapp/blob/master/js-miniapp-bridge/src/types/contact.ts) object as well.

js-miniapp-sample/src/pages/user-details.js
```diff
@@ -310,11 +310,27 @@ function UserDetails(props: UserDetailsProps) {
...
                 <ListItemAvatar>
                   <Avatar className={classes.contactIcon} />
                 </ListItemAvatar>
-                <ListItemText primary={contact.id} />
+                <ListItemText
+                  primary={contact.id}
+                  secondary={
+                    <React.Fragment>
+                      <Typography>
+                        {contact.name !== '' ? (
```
corycaywood, line 322: Will this be passed as an empty string when it doesn't exist? 

But `name` and `email` are optional properties, so I think they should be passed as `null` or `undefined` instead of an empty string.

ts-leojoseph-r: That's right. From iOS SDK, it is optional and it will pass as an empty string by default unless the Host app assigns `nil` to `name` and `email`.

Now, updated the code to not display the `name` & `email` if the value is `undefined` or `null`

minh-rakuten (approved):
👍 👍

js-miniapp-sample/src/pages/user-details.js
```diff
@@ -310,11 +308,31 @@ function UserDetails(props: UserDetailsProps) {
...
+                <ListItemText
+                  primary={contact.id}
+                  secondary={
+                    <React.Fragment>
+                      <Typography>
+                        {contact.name !== '' &&
+                        contact.name !== undefined &&
+                        contact.name !== null ? (
+                          <span>{'Name: ' + contact.name}</span>
+                        ) : null}
```
corycaywood, line 324: I think you don't need to manually check `undefined` and `null`. React treats the value `false` the same as `null` - it displays nothing.

```suggestion
                        {contact.name && contact.name !== '' && (
                          <span>{'Name: ' + contact.name}</span>
                        )}
```

corycaywood (approved):
LGTM 👍

PR #111, opened 2021-03-24
Audience and Scopes

Climbatize:
# Description
Adds Access Token scopes management

## Links

# Checklist
- [] I wrote/updated tests for new/changed code.


js-miniapp-bridge/src/types/token-data.ts
```diff
@@ -2,15 +2,35 @@
 export interface NativeTokenData {
   token: string;
   validUntil: number;
+  scopes?: NativeTokenPermission;
```
corycaywood, line 5: I think this isn't the correct return format, is it? We changed the iOS/Android SDK so that it returns `audience` and `scope` properties separately. i.e:
```
export interface NativeTokenData {
  token: string;
  validUntil: number;
  audience: string;
  scopes: string[];
}
```

js-miniapp-sdk/src/modules/user-info.ts
```diff
@@ -28,9 +28,14 @@ export interface UserInfoProvider {
...
   /**
    * Fetches the access token from host app.
+   * @param audience one of the audiences provided in MiniApp manifest
+   * @param scopes scopes array associated to the audience
    * @returns Access token from native host app.
    */
-  getAccessToken(): Promise<AccessTokenData>;
+  getAccessToken(
+    audience?: string,
+    scopes?: string[]
```
corycaywood, line 37: I think these properties shouldn't be optional. The Mini App is always required to pass these or else the iOS/Android SDK will return an error.

js-miniapp-bridge/src/types/token-data.ts
```diff
@@ -2,15 +2,35 @@
...
+}
+
+/** @internal */
+export interface NativeTokenPermission {
+  audience?: string;
+  scopes?: string[];
```
corycaywood, line 11: I think these properties shouldn't be optional. The iOS/Android side should always return a value.

js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -288,12 +288,14 @@ export class MiniAppBridge {
...
+   * @param {string} audienceRequested the audience the MiniApp requests for the token
+   * @param {string[]} scopesForAudience the associated scopes with the requested audience
    */
-  getAccessToken() {
+  getAccessToken(audienceRequested?: string, scopesForAudience?: string[]) {
     return new Promise<AccessTokenData>((resolve, reject) => {
       return this.executor.exec(
         'getAccessToken',
-        null,
+        { audience: audienceRequested, scopes: scopesForAudience },
```
corycaywood, line 298: Why use the property names `audienceRequested` and `scopesForAudience`? If you just use `audience` and `scopes`, you could simplify passing the data here because you can use shorthand property names:
```suggestion
        { audience, scopes },
```

js-miniapp-sdk/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - **Feature:** Support send message to contact. [See here](README.md#Send-message).
 - **Feature:** Added `Access Token` Custom permission.
 - **Feature:** Added optional `Name` & `Email` properties to Contact object
+- **Feature:** Added `Access Token` scopes and audience
```
corycaywood, line 8: Could you also update the access token section in the User Guide?

js-miniapp-bridge/src/types/token-data.ts
```diff
@@ -2,15 +2,35 @@
...
 }
 
 /** Token data type. */
 export class AccessTokenData {
   readonly token: string;
   readonly validUntil: Date;
+  readonly scopes: AccessTokenPermission;
```
corycaywood, line 18: See my above comment about the return format for audience/scopes. I think it will be easier to use if they can access these properties directly under `AccessTokenData`.

js-miniapp-bridge/src/types/token-data.ts
```diff
@@ -2,15 +2,35 @@
...
     this.token = baseToken.token;
     this.validUntil = new Date(baseToken.validUntil);
+    this.scopes = new AccessTokenPermission(baseToken.scopes);
+  }
+}
+
+/** Token permission type. */
+export class AccessTokenPermission {
+  readonly audience?: string;
+  readonly scopes?: string[];
```
corycaywood, line 30: See my other comment about these being optional: I think the SDK should always return these values.

corycaywood (commented):
Looking good! I left a few comments below.

js-miniapp-sdk/src/modules/user-info.ts
Climbatize: I'm just confused after discussing with @minh-rakuten, I just have to re-focus on this point. For now it's working with the current PR on iOS side, I really have to check with @minh-rakuten as it's the format I understood he also expected 😭

minh-rakuten: I think we can have `getAccessToken(): Promise<AccessTokenData>;` as deprecated function and the new one `getAccessToken(audience: string, scopes: string[]): Promise<AccessTokenData>;`

Climbatize: sorry that comment was for the previous feedback

js-miniapp-bridge/src/types/token-data.ts
Climbatize: I'm just confused after discussing with @minh-rakuten, I just have to re-focus on this point. For now it's working with the current PR on iOS side, I really have to check with @minh-rakuten as it's the format I understood he also expected 😭

corycaywood: Oh sorry I think I caused some confusion.

When I commented on the Android PR, I meant that it should change to match iOS with having audience and scope not be under a separate object.

But if what you have here is already working on both platforms, then I don't think we need to change it. Just probably the naming should be different:
- `NativeTokenScopes`
- `AccessTokenScopes`

minh-rakuten (approved):
Working well in android.
Thanks for getting this done 👏

minh-rakuten (commented):
You can add a unit test [here](https://github.com/rakutentech/js-miniapp/blob/master/js-miniapp-sdk/test/miniapp.spec.ts) for miniapp sdk when have time.

js-miniapp-sdk/src/modules/user-info.ts
Climbatize: I don't know @minh-rakuten. The old method will always fail because now scopes will be mandatory. That's a good notice

js-miniapp-bridge/src/types/token-data.ts
Climbatize: done

js-miniapp-bridge/src/types/token-data.ts
Climbatize: done

js-miniapp-bridge/src/common-bridge.ts
Climbatize: done

js-miniapp-sdk/CHANGELOG.md
Climbatize: yep

js-miniapp-bridge/src/types/token-data.ts
Climbatize: I think I did the necessary

Climbatize:
> You can add a unit test [here](https://github.com/rakutentech/js-miniapp/blob/master/js-miniapp-sdk/test/miniapp.spec.ts) for miniapp sdk when have time.

I did a small update, but I'll see if I can do better ;)

corycaywood (approved):
LGTM 👍

PR #112, opened 2021-03-30
Make contact name and e-mail fields optional

corycaywood:
# Description
Name and e-mail fields should be optional because they might not be set by the Host App.

# Checklist


ts-leojoseph-r (approved):
Looks good to me :+1:

minh-rakuten (approved):
👍

PR #113, opened 2021-04-01
fix: (sample) Remove 'delete' capability from file upload screen

corycaywood:
# Description
Having this capability is misleading as it's not possible to programmatically remove the file from the input. This was causing an issue where "deleting" the file would make it not possible to add the same file again. This was happening because the `setFile` function can only be called when input's `onChange` event occurs - but `onChange` doesn't get triggered if you select the same file which is already set to the input.

Also added the 'multiple' property to enable selecting multiple files.

## Links
MINI-3027

# Checklist


minh-rakuten (approved):
👍

PR #125, opened 2021-05-07
feat: send message permission

minh-rakuten:
# Description
- Add send message permission to bridge.
- Request send message permission before sending message in sample app.

## Links
MINI-3193

# Checklist

### DIFF
--- a/js-miniapp-bridge/src/types/custom-permissions.ts
+++ b/js-miniapp-bridge/src/types/custom-permissions.ts
@@ -4,7 +4,7 @@ export enum CustomPermissionName {
   CONTACT_LIST = 'rakuten.miniapp.user.CONTACT_LIST',
   ACCESS_TOKEN = 'rakuten.miniapp.user.ACCESS_TOKEN',
   LOCATION = 'rakuten.miniapp.device.LOCATION',
-  SEND_MESSAGE = 'rakuten.miniapp.action.SEND_MESSAGE',
+  SEND_MESSAGE = 'rakuten.miniapp.user.action.SEND_MESSAGE',
 }
 
 export enum CustomPermissionStatus {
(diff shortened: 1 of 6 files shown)


corycaywood (approved):
👍

js-miniapp-bridge/src/types/custom-permissions.ts
```diff
@@ -4,6 +4,7 @@ export enum CustomPermissionName {
   CONTACT_LIST = 'rakuten.miniapp.user.CONTACT_LIST',
   ACCESS_TOKEN = 'rakuten.miniapp.user.ACCESS_TOKEN',
   LOCATION = 'rakuten.miniapp.device.LOCATION',
+  SEND_MESSAGE = 'rakuten.miniapp.action.SEND_MESSAGE',
```
rleojoseph, line 7: @corycaywood Are we planning to use the same name in Android & iOS ?

As of now, in iOS it is `rakuten.miniapp.user.SEND_MESSAGE`, I was suggesting that we will change it to `rakuten.miniapp.user.action.SEND_MESSAGE`.

What do you think ?

corycaywood: Yeah, it should be the same on both platforms. I don't really have an opinion on which one we should use. @minh-rakuten What do you think?

minh-rakuten: We will use `rakuten.miniapp.user.action.SEND_MESSAGE`. cc @khairul-alam-licon

khairul-alam-licon: Ok, I'll update Android using `rakuten.miniapp.user.action.SEND_MESSAGE`

khairul-alam-licon (approved):
👍 👍

PR #129, opened 2021-05-12
fix: (sample) Changes to ad id inputs were not being saved to state

corycaywood:
# Description
The `.call` method is supposed to be passed the `this` context as the first parameter.

Also fixed a label issue on the "File Upload" screen - change "KB" to "Bytes".

# Checklist


khairul-alam-licon (approved):
👍 👍

Climbatize (approved):

PR #130, opened 2021-05-20
[MINI-3132] Update getAccessToken function to use a specific error type

Climbatize:
# Description
Update the `getAccessToken` function so that it receives specific error types. The Android/iOS side will send a specific error key for various cases. The JS SDK should then creates a `MiniAppError` subtype (extended from `Error`) specific to that error key and adds the appropriate message.

MiniApp sample 3.1.2

## Links
[MINI-3132](https://jira.rakuten-it.com/jira/browse/MINI-3132)

# Checklist


js-miniapp-sdk/src/miniapp-error-type.ts
```diff
@@ -0,0 +1,70 @@
...
+        messageArray.length > 0 ? messageArray.join(': ') : errorInput;
+    }
+  }
+}
+
+/**
+ * Enum for supported SDK error types
+ */
+export enum MiniAppErrorType {
+  AudienceNotSupportedError = 'AudienceNotSupportedError',
```
corycaywood, line 55: I was wondering if it would make more sense to make each type a custom error class? I saw something like this in the Firebase JS SDK:
`FirebaseError` (base error class): https://github.com/firebase/firebase-js-sdk/blob/5ad7ff2ae955c297556223e6cb3ad9d4b897f664/packages/util/src/errors.ts#L74
`FirebaseStorageError` (example error sub-class): https://github.com/firebase/firebase-js-sdk/blob/919413cd06478365e306476c8eb2ad2aa1f2d3ee/packages/storage/src/implementation/error.ts#L24

The advantage of this is that users can check the error type using `error instanceof FirebaseStorageError` and it also allows each error type to have custom properties.

So we could try something similar:
```
export class MiniAppErrorType extends Error {
  constructor(message: string) {
    super(message)
    this.name = "MiniAppErrorType"
    Object.setPrototypeOf(this, MyCustomError.prototype)
  }
}

export class AudienceNotSupportedError extends MiniAppErrorType {
  constructor() {
    super("The value passed for 'audience' is not supported.")
    this.name = "AudienceNotSupportedError"
    Object.setPrototypeOf(this, AudienceNotSupportedError.prototype)
  }
}
```

What do you think?

js-miniapp-sdk/src/miniapp-error-type.ts
```diff
@@ -0,0 +1,70 @@
+/**
+ * This class is a representation of an error sent from MiniApp mobile SDK
+ */
+export class MiniAppError extends Error {
```
corycaywood, line 4: Any custom classes and types that you want to be visible to users of the SDK need to be imported to and exported from `js-miniapp-sdk/src/index.ts`.

js-miniapp-sdk/README.md
```diff
@@ -411,6 +411,54 @@ miniApp.chatService.sendMessageToMultipleContacts(messageToContact)
...
+
+```javascript
+miniApp.user.getAccessToken("TOKEN_AUDIENCE", ["TOKEN_SCOPE1","TOKEN_SCOPE2"])
+  .then(data => {
+      ...
+  })
+  .catch(error => {
+      console.error(error.name);     // AudienceNotSupportedError
+      console.error(error.message); //  The value passed for 'audience' is not supported.
+      console.error(error.type);   //   AudienceNotSupportedError
```
corycaywood, line 440: I think we should have an example of how they would handle different types of errors. For example checking the `error.type` - or if you change it to use error classes as I mentioned in the other comment:
```
if (error instanceof AuthorizationFailureError) {
    // handle error
}
```

js-miniapp-sdk/src/miniapp-error-type.ts
Climbatize: I agree, this seems to be the standard way, I'll make an attempt with this in mind

js-miniapp-sdk/README.md
Climbatize: got it

js-miniapp-sdk/src/modules/user-info.ts
```diff
@@ -49,7 +60,36 @@ export class UserInfo implements UserInfoProvider {
...
-    return getBridge().getAccessToken(audience, scopes);
+  getAccessToken(
+    audience: string,
+    scopes: string[]
+  ): Promise<AccessTokenData | MiniAppError> {
+    return getBridge()
+      .getAccessToken(audience, scopes)
+      .catch(error => {
+        try {
+          const miniAppError = parseMiniAppError(error);
```
corycaywood, line 71: Shouldn't the types and parsing of the error be in the `js-miniapp-bridge` module? That way the parsing is bundled with the Android/iOS SDK and we don't need to worry about the Mini App having the an incompatible version of the JS SDK.

js-miniapp-sdk/src/miniapp-error-type.ts
```diff
@@ -0,0 +1,99 @@
...
+    this.customMessage = newMessage;
+  }
+
+  /**
+   * Takes the error input string sent from mobile SDK.
+   * Should be formatted as follow: "error_key: error message"
+   *
+   * @param errorInput error input string sent from mobile SDK
+   */
+  static fromCustomString(errorInput: string): MiniAppError {
```
corycaywood, line 38: See my message below about moving the parsing to the bridge module. In that case, I don't think there's a need to handle the case where the error is formatted as `key: message` since the parsing is bundled with the Android/iOS SDK.

js-miniapp-sdk/src/miniapp-error-type.ts
```diff
@@ -0,0 +1,99 @@
...
+ * This class is a representation of an error sent from MiniApp mobile SDK
+ */
+export class MiniAppError extends Error {
+  /**
+   * the raw input received from the SDK used to construct the MiniAppError
+   */
+  raw: string;
+  customMessage: string;
+
+  constructor(public errorInput: MiniAppJson) {
```
corycaywood, line 20: See here in the firebase repo: 
https://github.com/firebase/firebase-js-sdk/blob/5ad7ff2ae955c297556223e6cb3ad9d4b897f664/packages/util/src/errors.ts#L86

You have to set `Object.setPrototypeOf(this, MiniAppError.prototype)` or else I don't think you can use `instanceof` to check the type. I think it's some weird quirk of TypeScript.

Also, you need to set it in each sub-type: `Object.setPrototypeOf(this, AudienceNotSupportedError.prototype)`, etc.

corycaywood (commented):
It's looking good 👍  But as I mentioned, I think this needs to be moved to the `js-miniapp-bridge` package and then we can update the bridge module in the native SDKs.

js-miniapp-sdk/src/modules/user-info.ts
Climbatize: Yeah, that makes sense now you mention it :)

js-miniapp-sdk/src/miniapp-error-type.ts
Climbatize: I don't even use this method anymore, I will just remove it now we are in the bridge

.run/SDK build.run.xml
```diff
@@ -0,0 +1,15 @@
+<component name="ProjectRunConfigurationManager">
```
corycaywood, line 1: I think these are IDE configuration files? Probably you should add this folder to the `.gitignore`

js-miniapp-sdk/README.md
```diff
@@ -411,6 +411,60 @@ miniApp.chatService.sendMessageToMultipleContacts(messageToContact)
...
+Error messages sent to the bridge when a `getAccessToken` call is failing should always be in the following JSON format:
+
+````json
+{
+  "message": "error_message",
+  "type": "error_type_key"
+}
+````
+At the moment, the SDK will throw a `MiniAppError` (and subclasses), subclassing `Error` type, only on `getAccessToken` failures, based on `type` value.
+If the error `type` value is not supported, `Error.name` field will automatically be set to `Other`.
```
corycaywood, line 427: I think this seems like internal implementation details for the SDK - so maybe it's not needed here.

js-miniapp-bridge/src/types/error-types/index.ts
```diff
@@ -0,0 +1,76 @@
+/**
+ * Enum for supported SDK error types
+ */
+export enum MiniAppErrorType {
+  AudienceNotSupportedError = 'AudienceNotSupportedError',
+  ScopesNotSupportedError = 'ScopesNotSupportedError',
+  AuthorizationFailureError = 'AuthorizationFailureError',
+  Other = 'Other',
```
corycaywood, line 8: Is `Other` actually being used anywhere? It doesn't look like it's being used anymore, so maybe it can be removed.

js-miniapp-sdk/README.md
```diff
@@ -411,6 +411,60 @@ miniApp.chatService.sendMessageToMultipleContacts(messageToContact)
...
+
+Here are the `type` values currently supported and their default message:
+
+| Error Type | Message |
+| ---- | ---- |
+| `AudienceNotSupportedError` | `The value passed for 'audience' is not supported.` |
+| `ScopesNotSupportedError` | `The value passed for 'scopes' is not supported.` |
+| `AuthorizationFailureError` | _a message should be provided_ |
+| `Other` |  _a message should be provided_ |
```
corycaywood, line 436: This also seems like implementation details. Users will receive an instance of `AuthorizationFailureError`, so they shouldn't need to know the `type`, right?

corycaywood (approved):
Looks good 👍  Just looks like you need to update the User Guide again to remove the information from the old implementation.

js-miniapp-sdk/README.md
```diff
@@ -411,6 +411,60 @@ miniApp.chatService.sendMessageToMultipleContacts(messageToContact)
...
+
+You can retrieve the default message of a `MiniAppError` in the `Error.message` field. If a `message` value is provided, you can retrieve it in the `MiniAppError.customMessage` field
```
rleojoseph, line 438: `customMessage` is removed right ?

khairul-alam-licon (approved):
Looks good 👍

rleojoseph (approved):
Overall looks ok to me. Just a few comments from cory that need to be addressed.

js-miniapp-sdk/README.md
Climbatize: yep, rewrote it completely, thanks 👍

js-miniapp-sdk/README.md
Climbatize: You're right, the example is self-sufficient

js-miniapp-bridge/src/types/error-types/index.ts
Climbatize: it's true it's not, I remove it

js-miniapp-sdk/README.md
Climbatize: removed

PR #134, opened 2021-06-21
[MINI-3242] add interface to get user points

laubengaier:
# Description

First PR on JS SDK, any feedback would be nice!

Added interface to get user points.
- new method `getPoints` which returns a new `Points` type promise
- new permission `rakuten.miniapp.user.POINTS`

Simulator Screen Shot - iPhone 12 Pro - 2021-06-23 at 15 34 27

## Links
https://jira.rakuten-it.com/jira/browse/MINI-3242

# Checklist


js-miniapp-bridge/src/types/points.ts
```diff
@@ -0,0 +1,6 @@
+/** Points type. */
+export interface Points {
+  pointsStandard?: number;
+  pointsTerm?: number;
+  pointsCash?: number;
```
corycaywood, line 5: Maybe these could be shorted to just:
```suggestion
  standard?: number;
  term?: number;
  cash?: number;
```
What do you think?

js-miniapp-bridge/src/types/points.ts
```diff
@@ -0,0 +1,6 @@
+/** Points type. */
+export interface Points {
```
corycaywood, line 2: You also need to import this type to `js-miniapp-sdk/src/index.html` and then export it from that file so that the type is publicly accessible.

js-miniapp-sample/src/pages/tests/user-details.test.js
```diff
@@ -26,4 +28,14 @@ describe('Test Reducer', () => {
...
+
+  test('The points init action updates state properly', () => {
+    const newState = dataFetchReducer(initialState, { type: 'POINTS_FETCH_INIT' });
+    expect(newState.isLoading).toBe(true);
+  });
+
+  test('The points failure action updates state properly', () => {
+    const newState = dataFetchReducer(initialState, { type: 'POINTS_FETCH_FAILURE' });
+    expect(newState.isError).toBe(true);
+  });
```
corycaywood, line 40: We don't usually add unit tests for the JS Sample App because it's not used in production, so having tests can add unnecessary overhead.

Actually we should probably just remove all of the sample app tests sometime 🤔

js-miniapp-sample/src/pages/user-details.js
```diff
@@ -235,6 +267,38 @@ function UserDetails(props: UserDetailsProps) {
...
+    ];
+
+    props
+      .requestPermissions(permissionsList)
+      .then((permissions) =>
+        permissions
+          .filter(
+            (permission) => permission.status === CustomPermissionStatus.ALLOWED
+          )
+          .map((permission) => permission.name)
```
corycaywood, line 286: Looks like this is the same code used above to check for allowed permissions. So maybe you could extract it into a `filterAllowedPermissions` function?

js-miniapp-sample/src/pages/user-details.js
```diff
@@ -149,20 +153,46 @@ export const dataFetchReducer = (state: State, action: Action) => {
...
       };
+
+    case 'POINTS_FETCH_INIT':
+      return {
+        ...state,
+        isLoading: true,
+        isError: false,
+        hasRequestedPermissions: false,
+        hasRequestedPointPermissions: false,
+      };
```
corycaywood, line 180: I think you should create specific properties for points for all of these properties. Otherwise there could be some weird issues with errors showing in the wrong place. Also then you don't need to set `isLoading`, `isError`, etc because `...state` will bring in those properties if they already exist. For example:
```suggestion
      return {
        ...state,
        isPointsLoading: true,
        isPointsError: false,
        hasRequestedPointPermissions: false,
      };
```

js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -419,6 +420,21 @@ export class MiniAppBridge {
...
+   * Associating get point balance function to MiniAppBridge object.
+   * (provided rakuten.miniapp.user.POINTS is allowed by the user)
+   */
+  getPoints() {
+    return new Promise<Points>((resolve, reject) => {
+      return this.executor.exec(
+        'getPoints',
+        null,
+        points => resolve(JSON.parse(points) as Points),
+        error => reject(error)
```
corycaywood, line 434: I think for new functionality, we want to switch to using `MiniAppError` for errors instead of just passing a string. See `getAccessToken` above for an example. But I think in the case of `getPoints`, there aren't any specific error types which we want to support. So maybe just creating `MiniAppError` is enough?

@Climbatize @rleojoseph @khairul-alam-licon @munir-rakuten What do you think on this?

corycaywood (commented):
Looking good. I left a few comments below.

js-miniapp-sample/src/pages/tests/user-details.test.js
Climbatize: I agree on this @corycaywood

laubengaier: ok I've removed the two tests 😄

js-miniapp-bridge/src/common-bridge.ts
laubengaier: @corycaywood 👍 I've put default case with MiniAppError

```
...
switch (errorType) {
              default:
                return reject(new MiniAppError(miniAppError));
            }
```

corycaywood (approved):
LGTM 👍

Climbatize (approved):
LGTM

khairul-alam-licon (approved):
👍

PR #136, opened 2021-06-24
fixed the parsing conditions in getParams

munir-rakuten:
# Description
Parse Callback url properly when no parameters send from miniapp

## Links
MINI-2588

# Checklist


Climbatize (changes requested):
`?` is a valid character in query string, splitting on it can lead to an unwanted behavior

js-miniapp-sample/external-webview/index.html
```diff
@@ -85,12 +85,10 @@ <h3>Return To Mini App</h3>
...
             function getParam(name) {
-                const params = window.location.search;
-                if (params) {
-                    const param = params.substr(1).split("&")
-                        .map(param => param.split("="))
-                        .find(pair => pair[0] === name);
-                    return param && param[1];
+                const url = window.location.search;
+                if (url) {
+                    const params = new URL(url).searchParams;
```
corycaywood, line 90: I think this will throw an error, won't it? URL expects to receive a full URL. So maybe you could do this:
```suggestion
                    const params = new URL(window.location.href).searchParams;
```

munir-rakuten: @corycaywood i found that the url seems to be that
```
https://htmlpreview.github.io/?https://raw.githubusercontent.com/rakutentech/js-miniapp/master/js-miniapp-sample/external-webview/index.html
```
so there is a github preview url so to remove that and get the proper link i had to use `window.location.search` instead of `window.location.search`

corycaywood: Hmm, I see - but that makes this page have a dependency on the url always being in that format. For example, if we ever load the page from somewhere else where it isn't prepended with `https://htmlpreview.github.io/?`, then the page will not work.

Actually, line 108 is where it's supposed to parse out the `https://htmlpreview.github.io/?` portion. Maybe you can provide the URL that's parsed there instead of using `search`?

munir-rakuten: @corycaywood in line 108 its parsing the the params only..like using `window.location.search` removes the github preview and the replace regex removes the other url wth `?`

corycaywood (approved):
👍

khairul-alam-licon (approved):
👍👍

Climbatize (approved):
LGTM

PR #139, opened 2021-06-30
[gh-pages branch] fix: Minor versions > X.9 were not displaying in drop down menu

corycaywood:
# Description
The version 1.10 was not displaying in the drop down menu correctly because it was being interpreted as a number instead of string. I've added quotes so that it's a string.

Note that there is still an issue with the versions being in the wrong order (1.10 is displayed above 1.1) and the latest version redirect still redirects to 1.9. We will need to update the gh-pages template to fix this.

# Checklist


rleojoseph (approved):
Looks good to me :+1:

PR #142, opened 2021-07-09
[MINI-3452] Added a banner message

Climbatize:
# Description
Link has requested we support an optional message to the user, sent via the JS SDK when displaying the contact picker. 

## Acceptance:
Limit the text to 128 characters
Must accommodate English / Japanese characters and punctuation
Update the Demo JavaScript App to include this field as configurable for each of the three message types.

## Links
[MINI-3452](https://jira.rakuten-it.com/jira/browse/MINI-3452)

# Checklist


js-miniapp-bridge/src/types/message-to-contact.ts
```diff
@@ -8,4 +8,6 @@ export interface MessageToContact {
   caption: string;
   // Action which the call-to-action button will perform, i.e. a URI with parameters.
   action: string;
+  // Optional 128 characters message which can be used on contacts picker list for example
+  bannerMessage: string;
```
corycaywood, line 12: Including a message is optional, so this should be:
```suggestion
  bannerMessage?: string;
```

js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -361,7 +361,7 @@ export class MiniAppBridge {
     return new Promise<string | null>((resolve, reject) => {
       return this.executor.exec(
         'sendMessageToContact',
-        { messageToContact: message },
+        { messageToContact: messageTrimmingBannerText(message) },
```
corycaywood, line 364: I think it's a little strange to pass the whole message to this function when it only modifies one property. Maybe you could do something like this?
```suggestion
        { 
            messageToContact: {
                ...message,
                bannerMessage: trimBannerText(message.bannerMessage)
            }
         },
```

js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -464,3 +464,12 @@ function removeFromMessageQueue(queueObj) {
     mabMessageQueue.splice(messageObjIndex, 1);
   }
 }
+
+function messageTrimmingBannerText(message: MessageToContact, maxLength = 128) {
+  const hintString = message.bannerMessage.trim();
```
corycaywood, line 469: Since `bannerMessage` is optional, you need to handle that case to avoid an exception. Since we're using typescript, I think you could use optional chaining:
```suggestion
  const hintString = message.bannerMessage?.trim() ?? '';
```

corycaywood (commented):
Looking good 👍  I left a few comments below.

js-miniapp-bridge/src/common-bridge.ts
Climbatize: thanks for the suggestion @corycaywood !

rleojoseph (approved):
Looks good to me :+1:

khairul-alam-licon (approved):
Looks good 👍

js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -464,3 +480,9 @@ function removeFromMessageQueue(queueObj) {
     mabMessageQueue.splice(messageObjIndex, 1);
   }
 }
+
+function trimBannerText(message: string = undefined, maxLength = 128) {
```
corycaywood, line 484: @Climbatize @munir-rakuten Does the iOS/Android SDK support receiving `undefined`? We need to make sure this behaves as expected on the native side - i.e. The `bannerMessage` property should be optional in the native model.

corycaywood (approved):
👍

PR #145, opened 2021-07-14
[MINI-3476] removed banner field from single contact id screen

Climbatize:
# Description
QA was confused not seeing the banner message when sending a message to single contact id. I removed the banner message field so there is no more confusion

## Links
MINI-3476

# Checklist

### DIFF
--- a/js-miniapp-sample/src/pages/message.js
+++ b/js-miniapp-sample/src/pages/message.js
@@ -26,6 +26,12 @@ import {
 import { getMessageTypeList } from '../services/message/actions';
 import type { MessageType } from '../services/message/types';
 
+const MessageTypeId = {
+  SINGLE_CONTACT: 1,
+  SINGLE_CONTACT_ID: 2,
+  MULTIPLE_CONTACTS: 3,
+};
+
 const useStyles = makeStyles((theme) => ({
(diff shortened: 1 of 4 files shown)


js-miniapp-sample/src/pages/message.js
```diff
@@ -274,15 +273,17 @@ const Message = (props: MessageTypeProps) => {
...
-      <FormControl className={classes.formControl}>
-        <TextField
-          id="bannerMessage"
-          label="Banner message"
-          className={classes.fields}
-          onChange={onBannerMessageChange}
-          value={message.bannerMessage}
-        />
-      </FormControl>
+      {message.id !== 2 && (
```
corycaywood, line 276: We don't need to do this right now: but I think we should create constants for these ids so that this is easier to read.

corycaywood (approved):
Looks good 👍

khairul-alam-licon (approved):
👍

js-miniapp-sample/src/pages/message.js
Climbatize: yeah you're right

munir-rakuten (approved):

PR #147, opened 2021-07-19
(gh-pages) fix: Add date to versions to fix order of the list

corycaywood:
# Description
We had an issue where the versions are being sorted by name, but Jekyll cannot recognize it as semantic versions so the sorting is incorrect for versions greater than X.10. Jekyll will sort by date if the a date exists for each item, so I've added the date to each item to fix the sorting.

## Links
MINI-3425

# Checklist


Climbatize (approved):
Nice job

PR #148, opened 2021-07-19
fix: (sdk) Add date to docs version file to fix list order

corycaywood:
# Description
This fixes an issue where the versions in the documentation displayed in the wrong order when the version is greater than X.10

## Links
MINI-3425

# Checklist


Climbatize (approved):
You got it! It was not that hard to fix in the end :)

rleojoseph (approved):
Looks good to me :+1:

PR #150, opened 2021-07-21
sample app ui fix: user details and points show wrong UI info 

khairul-alam-licon:
# Description
Fix the wrong info when fetching user details and points. There  should be no dependency between these two features and wrong error message should not be seen if permission is already granted.

## Links
- MINI-3522

# Checklist

### DIFF
--- a/js-miniapp-sample/src/pages/user-details.js
+++ b/js-miniapp-sample/src/pages/user-details.js
@@ -157,7 +157,6 @@ export const dataFetchReducer = (state: State, action: Action) => {
         isLoading: true,
         isError: false,
         hasRequestedPermissions: false,
-        hasRequestedPointPermissions: false,
       };
     case 'FETCH_SUCCESS':
       return {
@@ -165,7 +164,6 @@ export const dataFetchReducer = (state: State, action: Action) => {
         isLoading: false,
         isError: false,
         hasRequestedPermissions: true,
-        hasRequestedPointPermissions: false,
       };
     case 'FETCH_FAILURE':
       return {
@@ -179,7 +177,6 @@ export const dataFetchReducer = (state: State, action: Action) => {
         ...state,
         isPointsLoading: true,
         isPointsError: false,
-        hasRequestedPermissions: false,
         hasRequestedPointPermissions: false,
       };
     case 'POINTS_FETCH_SUCCESS':
@@ -187,7 +184,6 @@ export const dataFetchReducer = (state: State, action: Action) => {
         ...state,
         isPointsLoading: false,
         isPointsError: false,
-        hasRequestedPermissions: false,
         hasRequestedPointPermissions: true,
       };
     case 'POINTS_FETCH_FAILURE':
--- a/js-miniapp-sample/src/pages/user-details.js
+++ b/js-miniapp-sample/src/pages/user-details.js
@@ -157,8 +157,6 @@ export const dataFetchReducer = (state: State, action: Action) => {
         isLoading: true,
         isError: false,
         hasRequestedPermissions: false,
-        isPointsLoading: false,
-        isPointsError: false,
         hasRequestedPointPermissions: false,
       };
     case 'FETCH_SUCCESS':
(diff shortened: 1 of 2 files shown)


js-miniapp-sample/src/pages/user-details.js
```diff
@@ -157,40 +157,56 @@ export const dataFetchReducer = (state: State, action: Action) => {
         isLoading: true,
         isError: false,
         hasRequestedPermissions: false,
+        isPointsLoading: false,
+        isPointsError: false,
+        hasRequestedPointPermissions: false,
```
corycaywood, line 162: I think it's incorrect to add all of the state items here: this will reset state items which the action had no affect on. For example, if points is already loading or has an error, then this will incorrectly reset it to `false`. 
The `...state` line above is already picking the correct values for these state items.

corycaywood (changes requested):

js-miniapp-sample/src/pages/user-details.js
khairul-alam-licon: Ok, I have updated the PR to reset only the permission item state during INIT & SUCCESS operations.

js-miniapp-sample/src/pages/user-details.js
```diff
@@ -157,13 +157,15 @@ export const dataFetchReducer = (state: State, action: Action) => {
         isLoading: true,
         isError: false,
         hasRequestedPermissions: false,
+        hasRequestedPointPermissions: false,
```
corycaywood, line 160: I think it's still incorrect to set `hasRequestedPointPermissions` under `FETCH_INIT`, `FETCH_SUCCESS`, `FETCH_FAILURE` because these are for fetching the other user details, not the points details.

js-miniapp-sample/src/pages/user-details.js
```diff
@@ -177,20 +179,22 @@ export const dataFetchReducer = (state: State, action: Action) => {
         ...state,
         isPointsLoading: true,
         isPointsError: false,
+        hasRequestedPermissions: false,
```
corycaywood, line 182: It's also incorrect to set `hasRequestedPermissions` under `POINTS_FETCH_INIT`, `POINTS_FETCH_SUCCESS`, and `POINTS_FETCH_FAILURE` since these are the actions for points and not for user details.

corycaywood (commented):
I think this still isn't correct because there are state items being updated which aren't related to the action. Shouldn't just the update to `isPointsError` be enough to fix the issue?

js-miniapp-sample/src/pages/user-details.js
khairul-alam-licon: @corycaywood Initially, I thought only updating to `isPointsError` should enough to fix the issue.
But, I have rechecked that we also need to set `hasRequestedPointPermissions: false` in here to fix the issue.

js-miniapp-sample/src/pages/user-details.js
khairul-alam-licon: I have rechecked that we also need to set `hasRequestedPermissions: false` in here to fix the issue.

js-miniapp-sample/src/pages/user-details.js
corycaywood: @khairul-alam-licon What issue does it fix to have `hasRequestedPointPermissions` here? I can see that you need to set `hasRequestedPointPermissions: true` under the `POINTS_FETCH_SUCCESS` action.

But setting this under `FETCH_INIT` and `FETCH_SUCCESS` will still cause incorrect behavior: these actions are for the user details and are unrelated to the Points functionality so they should not modify any of the Points state. For example, if you tap "Fetch points" and then deny the points permission, you will correctly get the error state for Points. But currently if you were then to tap "Fetch user details" then it would reset the points error state.

corycaywood (approved):
Looks good 👍

js-miniapp-sample/src/pages/user-details.js
khairul-alam-licon: @corycaywood Yeah, as we discussed and you got the root cause, it had issues with fetching all the permissions by https://github.com/rakutentech/js-miniapp/blob/master/js-miniapp-sample/src/services/permissions/reducers.js 
Thank you for your suggestion 🙇

js-miniapp-sample/src/pages/user-details.js
khairul-alam-licon: Updated per suggestion.

rleojoseph (approved):
Looks good to me :+1:

PR #151, opened 2021-08-05
changelog for release 1.11.0

Climbatize:


rleojoseph (approved):
Looks good to me :+1:

js-miniapp-sdk/CHANGELOG.md
```diff
@@ -1,4 +1,13 @@
 ## CHANGELOG
+### 1.11.0 (2021-08-05)
+- **Feature:** Added interface to get user points
+- **Fix:** fixed the parsing conditions in sample app
+- **Fix:** added checking for auth token error
+- **Sample:** Added Buttons to select different file types
+- **Feature:** Added an optional banner message field in messaging feature
+- **Sample:** Added callback url field to URI screen so it can be changed if the mini app is being used in a different Host App
```
corycaywood, line 8: We haven't previously been putting items for the Sample in this changelog. I don't think it makes sense to add it here because it's not tied to the JS SDK version . Also the JS Sample isn't really used by the JS SDK users the same way that the Android/iOS demo app is used.

If we want to have a changelog for the Sample App, I think it should be a separate one in the `js-miniapp-sample` folder.

corycaywood (approved):
👍

khairul-alam-licon (approved):
👍

PR #156, opened 2021-09-16
[MALI-604] Expose Host Environment Information from Miniapp

laubengaier:
# Description
Exposing the Host Environment Information from MiniApps. Platform is used and attached with the prior method all other version infos should be retrieved through the Hosts implementation.

https://user-images.githubusercontent.com/17570451/133572021-91c94727-5650-4a6e-80bd-cadc45f3c576.mov

## Links
https://jira.rakuten-it.com/jira/browse/MALI-604

# Checklist


js-miniapp-bridge/src/index.ts
```diff
@@ -44,6 +46,8 @@ export {
   AccessTokenData,
   Contact,
   Points,
+  HostEnvironmentInfo,
+  Platform,
```
Climbatize, line 50: Isn't `Platform` part of `HostEnvironmentInfo`?

js-miniapp-sdk/src/miniapp.ts
```diff
@@ -203,4 +211,13 @@ export class MiniApp implements MiniAppFeatures, Ad, Platform {
   getPoints(): Promise<Points> {
     return getBridge().getPoints();
   }
+
+  getHostEnvironmentInfo(): Promise<HostEnvironmentInfo> {
+    return getBridge()
+      .getHostEnvironmentInfo()
+      .then(info => {
+        info.platform = HostPlatform[getBridge().platform.toUpperCase()];
```
corycaywood, line 219: I think you don't need to add `toUpperCase`. That is making it so the value for `platform` doesn't match what's in the `Platform` type.

Probably the text should be either all uppercase or all lowercase, but I think it's too late to change it since we've already been exposing it through the `MiniApp.getPlatform` interface.

corycaywood (approved):
Looks good 👍  Just see one comment below.

js-miniapp-sdk/src/miniapp.ts
corycaywood: Oh wait, sorry you can ignore my comment. I see now that this is uppercased just for getting the type via enum.

But maybe you could use something like this instead? `getBridge().platform as HostPlatform`

khairul-alam-licon (approved):
Looks good 👍

js-miniapp-bridge/src/index.ts
laubengaier: @Climbatize the Platform is a model itself it's only imported into HostEnvironmentInfo. not totally sure but that needs to be exported as well I think?

```
-> platform.ts
/** @internal */

/** Device platform. */
export enum Platform {
  ANDROID = 'Android',
  IOS = 'iOS',
}
```

```
import { Platform } from './platform';

/** HostEnvironmentInfo type. */
export interface HostEnvironmentInfo {
  platform?: Platform;
  platformVersion?: string;
  hostVersion?: string;
  sdkVersion?: string;
}
```

js-miniapp-sdk/src/miniapp.ts
laubengaier: @corycaywood somehow didn't work for me when testing at first but somehow does now so I guess that's better :D

Climbatize (approved):

PR #158, opened 2021-09-30
MINI-4074 Handle external webview close event

munir-rakuten:
# Description
Handle external webview close event.
Screen Shot 2021-10-11 at 16 28 07

## Links
MINI-4074

# Checklist


js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -482,6 +482,17 @@ export class MiniAppBridge {
       );
     });
   }
+
+  setSecondaryWebviewClose(): Promise<string> {
+    return new Promise<string>((resolve, reject) => {
```
corycaywood, line 487: I don't think a Promise is the right way to handle this. A promise is used for a one-time asynchronous operation which can either succeed or fail. However "closing the secondary webview" is an event (not an operation) that could occur multiple times in the lifetime of a mini app and doesn't have a failure state.

So instead, I think we should either:

1. Let them pass a function which we save and then call when the JS SDK receives the "close" event. i.e. `setWebviewClosedListener( (message) => { console.log("webview closed") })`. This would probably work similar to how we have setup the `exec` function - so we might need to create a separate bridge function on the native side for passing this event.
2. OR Create a [synthetic event](https://developer.mozilla.org/en-US/docs/Web/Events/Creating_and_triggering_events) that the Mini App can listen for such as `window.addEventListener("miniAppWebviewClosed", () => {} )`.

I think option 2 would require us to also implement an interface similar to optional 1 anyways, so maybe just doing 1 is okay.

What do you think?

corycaywood (commented):
Thanks for investigating this, it looks like a good start. See my comment below: I think we need to handle this as a listener function instead of as a Promise.

js-miniapp-sample/src/pages/uri-schemes.js
```diff
@@ -41,6 +42,16 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
+const requestSecondaryWebviewClose = () => {
+  MiniApp.setSecondaryWebviewClose()
+    .then((message) => {
+      console.log(message);
+    })
+    .catch((error) => {
+      console.error(error);
```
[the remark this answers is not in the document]
munir-rakuten: @corycaywood yes,  that error catch may not be used anytime, but still put it to make the code consistent with the other implementations.

js-miniapp-bridge/src/common-bridge.ts
munir-rakuten: @corycaywood i will look into it, thanks for the suggestions

js-miniapp-sample/src/pages/landing.js
```diff
@@ -13,6 +13,16 @@ type LandingProps = {
   getHostInfo: Function,
 };
 
+window.addEventListener('miniAppWebviewClosed', function (e) {
+  let message = e.detail;
+  console.log(message);
+});
+
+window.addEventListener('miniAppOnPause', function (e) {
```
corycaywood, line 21: I think we also need a `miniappresume` event. I know that we have `miniAppWebviewClosed` but that's specifically for use when the webview was closed. There are other cases where the Host App could pause the mini app when they wouldn't get a `miniAppWebviewClosed` - i.e. putting the app in the background, receive a phone call, etc.

corycaywood, line 21: I think these should be `miniappwebviewclosed` and `miniapppause` instead. It's not as easy to read, but it's the standard format used by JavaScript events (see https://developer.mozilla.org/en-US/docs/Web/Events)

js-miniapp-sample/src/pages/landing.js
```diff
@@ -13,6 +13,16 @@ type LandingProps = {
   getHostInfo: Function,
 };
 
+window.addEventListener('miniAppWebviewClosed', function (e) {
```
corycaywood, line 16: Maybe we should have a separate screen in the sample mini app for testing the events? Currently the only way to open an external webview is on the "Uri Schemes" screen.

For example, the "event listener" screen could have:
- Button to open a page in the external webview
- Counter showing how many times it has received the `miniappwebviewclosed ` event 
- Counter showing how many times received `miniapppause` event
- Counter showing how many times received `miniappresume` event

What do you think?

corycaywood (commented):
Looks good! I left a few more suggestions below.

js-miniapp-sample/src/pages/event-listener.js
```diff
@@ -0,0 +1,118 @@
...
+  }
+});
+
+window.addEventListener(MiniAppEvents.pause, function (e) {
+  let message = e.detail;
+  console.log(message);
+  pauseeventcount++;
+  const pause = document.getElementById('pause');
+  if (pause) {
+    pause.innerText = 'Mini App Paused: ' + pauseeventcount;
```
corycaywood, line 68: These values should be updated reactively. I think that's why you're getting errors on CI: you aren't really supposed to directly manipulate the DOM when using React.

You can use state hooks for this, see some of the other screens for an example, i.e. https://github.com/rakutentech/js-miniapp/blob/master/js-miniapp-sample/src/pages/local-storage.js#L48

So in your case, you would do something like:
```
const [pauseEventCount, setPauseEventCount] = useState(0);

window.addEventListener(MiniAppEvents.pause, function (e) {
  setPauseEventCount(pauseeventcount++);
}

// And use directly in the JSX
<p>Mini App Paused: {pauseEventCount}</p>
```

js-miniapp-sample/src/pages/event-listener.js
```diff
@@ -0,0 +1,118 @@
...
+    wordBreak: 'break-all',
+    color: theme.color.primary,
+    marginTop: 0,
+    paddingBottom: 10,
+  },
+}));
+
+let externalwebviewcloseeventcount = 0;
+let pauseeventcount = 0;
+let resumeeventcount = 0;
```
corycaywood, line 49: I think these can use camel case - only the event name strings are supposed to be all lowercase.

js-miniapp-sdk/src/event-types/index.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Enum for supported SDK event types
+ */
+export enum MiniAppEvents {
+  externalwebviewclose = 'miniappwebviewclosed',
+  pause = 'miniapppause',
+  resume = 'miniappresume',
```
corycaywood, line 7: Maybe the enum names should use all caps with snake case? It's the format we used for most other enums in the SDK:
```suggestion
  EXTERNAL_WEBVIEW_CLOSE = 'miniappwebviewclosed',
  PAUSE = 'miniapppause',
  RESUME = 'miniappresume',
```

js-miniapp-sample/src/pages/event-listener.js
```diff
@@ -0,0 +1,118 @@
...
+
+window.addEventListener(MiniAppEvents.externalwebviewclose, function (e) {
+  let message = e.detail;
+  console.log(message);
+  externalwebviewcloseeventcount++;
+  const webviewclose = document.getElementById('webviewclose');
```
corycaywood, line 55: See my other comment about reactively updating the variable. `getElementById` typically shouldn't need to be used in a React app.

js-miniapp-sample/src/pages/event-listener.js
```diff
@@ -0,0 +1,118 @@
...
+  if (resume) {
+    resume.innerText = 'Mini App Resumed: ' + resumeeventcount;
+  }
+});
+
+const NativeEvents = () => {
+  const classes = useStyles();
+
+  function onOpenExternalWebview() {
+    const EXTERNAL_WEBVIEW_URL = 'https://www.google.com';
```
corycaywood, line 86: Maybe the constant should be moved near the top of this file?

PR #159, opened 2021-10-12
fix: (sample) Allow messaging screen to be scrollable

corycaywood:
# Description
Previously the text inputs were not visible on Android when opening the keyboard, because the keyboard overlayed on top of them.

I'm not sure what's wrong with the diff on `message.js`. Actually the only change was that I added `<div className={classes.scrollable}>` to surround the `<Fragment>`, but for some reason it's calculating the diff really strange.

## Links
MALI-679

# Checklist


khairul-alam-licon (approved):
Looks good 👍

PR #163, opened 2021-10-19
Improvement for value update in custom events

munir-rakuten:
# Description
As there is no way to update the detail read only property of a custom event so to update the value of the same event type
first removed the event from the queue and then add a new one with the same event type.

## Links

# Checklist


corycaywood (approved):
Looks good 👍

khairul-alam-licon (approved):
Looks good 👍

PR #181, opened 2021-12-03
[MINI-4423] Take picture with camera fix

laubengaier:
# Description
Fix for the camera feature not displaying the file name

# Checklist

### DIFF
--- a/js-miniapp-sample/src/pages/camera.js
+++ b/js-miniapp-sample/src/pages/camera.js
@@ -1,11 +1,6 @@
 import React, { useState } from 'react';
 
-import {
-  Card,
-  Grid,
-  Button,
-  makeStyles,
-} from '@material-ui/core';
+import { Card, Grid, Button, makeStyles } from '@material-ui/core';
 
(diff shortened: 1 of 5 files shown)


js-miniapp-sample/src/pages/camera.js
```diff
@@ -87,6 +90,7 @@ const Camera = () => {
             data-testid="file-input-image-back"
             capture="environment"
             value={backCamera}
+            ref={cameraRef}
```
corycaywood, line 93: I think the problem might be that you don't need to set `value={backCamera}` at all. The `input` should get the value set automatically by the browser after a file is chosen. At least that seems to be how it's working on the "File upload" screen:
https://github.com/rakutentech/js-miniapp/blob/master/js-miniapp-sample/src/pages/file-upload.js#L93

laubengaier: thanks but yeah got it working in the end 😄

Climbatize (approved):

corycaywood (approved):
👍

PR #182, opened 2021-12-07
[MINI-1733] Feat/purchase item

rleojoseph:
# Description
* Updated SDK, Bridge & Sample app to use Purchase item interface implementation.

## Links
MINI-1733

# Checklist


js-miniapp-bridge/src/types/purchaseProduct.ts
```diff
@@ -0,0 +1,19 @@
...
+  id: string;
+  title: string;
+  description: string;
+  price: ProductPrice;
+}
+
+export interface ProductPrice {
+  amount: number;
+  currencyCode: string;
+  price: string;
```
khairul-alam-licon, line 12: In Android, there is an object e.g. [SkuDetails](https://developer.android.com/reference/com/android/billingclient/api/SkuDetails) returns price as micro, if iOS has a similar thing, we can add a new field for that in `ProductPrice`.

khairul-alam-licon (approved):
Looks good to me 👍

Climbatize (approved):

PR #188, opened 2022-02-04
sample:  Change label on deep link field

corycaywood:
# Description
The deep link field can also be used to open any external website, so updated the title to reflect that. Also some small improvements to use react hooks for setting the field state and fixed an issue to prevent the webview from redirecting when the URL is empty.

# Checklist


khairul-alam-licon (approved):
Looks good 👍

rleojoseph (approved):
Looks good to me :+1:

PR #194, opened 2022-02-21
[MINI-4710] adding keyboard events

laubengaier:
# Description
Adding keyboard events for CE for showing and hiding the keyboard
- currently passing the screenHeight and keyboardHeight
- setting the height to zero when the keyboard is hidden

https://user-images.githubusercontent.com/17570451/154896257-4cdb64c1-50c2-45d3-ba9e-46a8bdfa5888.MP4

## Links
https://jira.rakuten-it.com/jira/browse/MINI-4710

# Checklist


js-miniapp-sdk/src/event-types/index.ts
```diff
@@ -5,4 +5,6 @@ export enum MiniAppEvents {
   EXTERNAL_WEBVIEW_CLOSE = 'miniappwebviewclosed',
   PAUSE = 'miniapppause',
   RESUME = 'miniappresume',
+  KEYBOARDSHOWN = 'keyboardshown',
+  KEYBOARDHIDDEN = 'keyboardhidden',
```
corycaywood, line 9: I think these should be pre-fixed with `miniapp` to match the other events and prevent conflicting event names in the mini app:
```suggestion
  KEYBOARDSHOWN = 'miniappkeyboardshown',
  KEYBOARDHIDDEN = 'miniappkeyboardhidden',
```

js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -121,9 +121,25 @@ export class MiniAppBridge {
...
+    navigationBarHeight: number,
+    screenHeight: number,
+    keyboardHeight: number
+  ) {
+    const event = new CustomEvent(eventType, {
+      detail: {
+        message: value,
+        navigationBarHeight,
+        screenHeight,
+        keyboardHeight,
```
corycaywood, line 140: Isn't this function and `CustomEvent` also used for the `resume` and `pause` events? So this would mean that those events always contain `height` properties for all of the events but they're `null` in the case of `pause`, `resume`, etc.

I think it would be more clear to the SDK users if we have a specific type for the keyboard events, i.e. `MiniAppKeyboardEvent` or something. I'm not sure the best way to handle this, but maybe we need a separate function for `execKeyboardEvent`?

corycaywood (commented):
Looking good - I left a couple of suggestions below.

js-miniapp-sdk/src/event-types/index.ts
laubengaier: fixed

js-miniapp-bridge/src/common-bridge.ts
laubengaier: @corycaywood I fixed it just I logged the `execCustomEventsCallback` & `execKeyboardEventsCallback` and saw that the logs duplicates with every call is that something we have to take a deeper look into?

corycaywood: @laubengaier Do you mean that the callback for `execCustomEventsCallback` was getting called twice even before the keyboard callback was added?

It does seem like it would be a problem, because the mini app's callback would be invoked twice.

Climbatize: @laubengaier where did you spot this log? on iOS console in Sample App project?

laubengaier: @corycaywood @Climbatize 

I've already added an extra queue for the keyboards events but it seems to trigger it js side a bunch of times. I've changed the bridge code inside the iOS project to log the keyboard event like below:

```
     MiniAppBridge.prototype.execKeyboardEventsCallback = function (eventType, message, navigationBarHeight, screenHeight, keyboardHeight) {
         console.log('### keyboard event: ' + eventType);
         var event = new CustomEvent(eventType, {
             detail: {
                 message: message,
                 navigationBarHeight: navigationBarHeight,
                 screenHeight: screenHeight,
                 keyboardHeight: keyboardHeight,
             },
         });
         var queueObj = mabKeyboardEventQueue.filter(function (customEvent) { return customEvent === event; })[0];
         if (!queueObj) {
             
             if (eventType === event.type) {
                 console.log('### keyboard remove event: ' + event.type);
                 removeFromKeyboardEventQueue(event);
             }
             queueObj = event;
             mabKeyboardEventQueue.unshift(queueObj);
         }
         this.executor.execEvents(queueObj);
     };
```
ends up looking like this:
Screen Shot 2022-03-14 at 12 10 58

not too sure about queues in js

corycaywood: @laubengaier Seems like it must be an issue with the JS Sample App. Actually, `removeEventListener` is never called by the Sample App, so it is probably being attached multiple times without removing the old one. Probably we should hook into the React component state to add/remove the listeners.

I think we can address that later in another PR, so this one is okay to merge for now.

corycaywood (approved):
👍

rleojoseph (approved):
Looks good to me :+1:

PR #207, opened 2022-04-04
[NO-TICKET] Folder with localized manifests

laubengaier:
# Description
Adding folder with localized manifests

# Checklist

### DIFF
--- a/js-miniapp-sample/public/_locale/ja/manifest.json
+++ b/js-miniapp-sample/public/_locale/ja/manifest.json
@@ -1,0 +1,73 @@
+{
+  "short_name": "Mini App",
+  "name": "JS MiniApp Sample",
+  "icons": [
+    {
+      "src": "favicon.ico",
+      "sizes": "64x64 32x32 24x24 16x16",
+      "type": "image/x-icon"
+    }
+  ],
+  "start_url": ".",
+  "display": "standalone",
+  "theme_color": "#000000",
+  "background_color": "#ffffff",
+
+  "reqPermissions": [
+    {
+      "name": "rakuten.miniapp.user.USER_NAME",
+      "reason": "Describe your reason here (optional)."
+    },
+    {
+      "name": "rakuten.miniapp.user.PROFILE_PHOTO",
+      "reason": "Describe your reason here (optional)."
+    }
+  ],
+
+  "optPermissions": [
+    {
+      "name": "rakuten.miniapp.user.CONTACT_LIST",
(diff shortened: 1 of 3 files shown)


corycaywood (commented):
@laubengaier This format for the manifest works, but I think it would be better to use the placeholder format. That way we don't have to duplicate content across the files:

```

// manifest.json:
{
    "customMetaData": {
        "exampleKey": "__MANIFEST_exampleText__",
    }
}
 
// _locales/default/manifest.json:
{
  "exampleText": "Default text."
}
 
// _locales/ja/manifest.json:
{
  "exampleText": "Japanese text."
}
```

js-miniapp-sample/public/_locale/ja/manifest.json
```diff
@@ -0,0 +1,73 @@
...
+    {
+      "name": "rakuten.miniapp.user.PROFILE_PHOTO",
+      "reason": "Describe your reason here (optional)."
+    }
+  ],
+
+  "optPermissions": [
+    {
+      "name": "rakuten.miniapp.user.CONTACT_LIST",
+      "reason": "Describe your reason here (optional)."
```
corycaywood, line 30: Maybe we should add a translation for the `reason` too?
I google translated this 😆 : `ここに理由を説明してください（オプション）。`

rleojoseph (changes requested):
@laubengaier 
`manifest.json` from the main folder should contain the key for these localized texts
```

  {
      "name": "rakuten.miniapp.user.USER_NAME",
      "reason": "__MANIFEST_userNameReason__"
  },
  {
      "name": "rakuten.miniapp.user.POINTS",
      "reason": "__MANIFEST_pointsReason__"
  }
```

And the `manifest.json` inside _locale folders will contain only the value of those keys, like

_locale/default/manifest.json

```
{
  "userNameReason": "Default text.",
  "pointsReason: "Default text"
}
```

_locale/ja/manifest.json

```
{
  "userNameReason": "JP Default text.",
  "pointsReason: "JP Default text"
}
```

js-miniapp-sample/public/_locale/ja/manifest.json
laubengaier: :D yeah I think that works (anyways demo purpose)

rleojoseph (approved):
Looks good to me :+1:

corycaywood (approved):
Looks good 👍

PR #216, opened 2022-04-26
feat: (bridge) Add error types for File Download

corycaywood:
# Description
Added error type parsing for the Filed Download errors. In particular we need the `InvalidUrlError` so developers know when they've provided an unsupported type (only supports https, http, and data:).
Also refactored the MiniAppError type to be easier to re-use for different features

## Links
MINI-4979

# Checklist
- [x] I have completed all steps in the [Contribution Checklist](../.github/CONTRIBUTING.md#checklist)


khairul-alam-licon (approved):
Looks good 👍

rleojoseph (approved):
Thanks for this refactoring. This is much better. Looks good to me :+1:

PR #220, opened 2022-05-10
fix: Secure storage ready/error events not received in some cases

corycaywood:
# Description
The ready/error events are sent very early when the mini app starts, so the mini app sometimes calls the event listener after the event has already been received. To handle this case, I added a variable in the bridge to cache the state. Also changed the SDK interfaces to accept callback functions instead of Promises

# Checklist
- [x] I have completed all steps in the [Contribution Checklist](../.github/CONTRIBUTING.md#checklist)


rleojoseph (approved):
Looks good to me :+1:

khairul-alam-licon (approved):
Looks good 👍

PR #258, opened 2023-06-23
[MINI-6189] - Send Analytics to Miniapp

rleojoseph:
# Description
* Miniapp can send analytics using this new interface

## Links
[MINI-6189](https://jira.rakuten-it.com/jira/browse/MINI-6189)

# Checklist

### DIFF
--- a/js-miniapp-bridge/src/common-bridge.ts
+++ b/js-miniapp-bridge/src/common-bridge.ts
@@ -832,6 +832,6 @@ function BooleanValue(value) {
     } else if (lowerCaseValue === 'false' || lowerCaseValue === '0') {
       return false;
     }
-    return false;
   }
+  return false;
 }
--- a/js-miniapp-bridge/src/common-bridge.ts
+++ b/js-miniapp-bridge/src/common-bridge.ts
@@ -29,6 +29,7 @@ import { MiniAppError, parseMiniAppError } from './types/error-types';
 import { MiniAppResponseInfo } from './types/response-types/miniapp';
 import { ProductInfo, PurchasedProductInfo } from './types/in-app-purchase';
 import { HostThemeColor } from './types/host-color-scheme';
+import { MAAnalytics } from './types/analytics/analytics';
 
 /** @internal */
 const mabMessageQueue: Callback[] = [];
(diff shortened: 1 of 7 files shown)


GondiTilak (approved):
Looks good to me :+1:

js-miniapp-bridge/src/common-bridge.ts
```diff
@@ -811,3 +821,10 @@ function trimBannerText(message: string = null, maxLength = 128) {
     ? message?.substring(0, maxLength - 1) + '…'
     : message;
 }
+
+function BooleanValue(value) {
+  if (value.toLowerCase() === 'true') {
+    return true;
+  }
+  return false;
+}
```
Climbatize, line 830: I would suggest a more flexible function:

```suggestion
function BooleanValue(value) {
  if (typeof value === 'boolean') {
    // If the value is already a boolean, return it
    return value;
  } else if (typeof value === 'string') {
    // If the value is a string, check for common representations of boolean values
    const lowerCaseValue = value.toLowerCase();
    if (lowerCaseValue === 'true' || lowerCaseValue === '1') {
      return true;
    } else if (lowerCaseValue === 'false' || lowerCaseValue === '0') {
      return false;
    }
  } else if (typeof value === 'number') {
    // If the value is a number, check for integer representations of boolean values
    if (value === 1) {
      return true;
    } else if (value === 0) {
      return false;
    }
  }

  // If the value cannot be interpreted as a boolean, return false
  return false;
}
```

unit tests:
```javascript
console.log(BooleanValue(true)); // true
console.log(BooleanValue(false)); // false
console.log(BooleanValue('true')); // true
console.log(BooleanValue('false')); // false
console.log(BooleanValue('1')); // true
console.log(BooleanValue('0')); // false
console.log(BooleanValue(1)); // true
console.log(BooleanValue(0)); // false
console.log(BooleanValue(null)); // false
console.log(BooleanValue('hello')); // false
console.log(BooleanValue(42)); // false
```

Climbatize (approved):
LGTM. See my suggestion

js-miniapp-bridge/src/common-bridge.ts
rleojoseph: Thanks for the feedback. We don't need to take care of number types though, I have updated the function now

munir-rakuten (commented):
looks good to me, thanks.

PR #284, opened 2023-11-14
[MALI-1382] Contacts encoding

rleojoseph:
# Description
If any of the contact names is having any special characters like ' or '' , it would break the entire contact list. Hence, we are providing an option for the native app to send the encoded contact names with the updated interface

## Links
https://jira.rakuten-it.com/jira/browse/MALI-1382

# Checklist


GondiTilak (approved):
Looks good to me :+1:

munir-rakuten (approved):
Looks good to me.

PR #317, opened 2025-06-03
OAF-26391 - add launch app settings interface

Toby-SingLeungHo:
# Description
Add Launch App Settings interface to bridge, sdk and create sample page

## Links
[OAF-26391](https://jira.rakuten-it.com/jira/browse/OAF-26391)

# Checklist

### DIFF
--- a/js-miniapp-bridge/src/modules/utility-manager.ts
+++ b/js-miniapp-bridge/src/modules/utility-manager.ts
@@ -46,4 +46,19 @@ export class UtilityManager {
       );
     });
   }
+
+  /**
+   * Trigger launchAppSettings from host
+   * @returns true or false whether launch app settings is enabled or not
+   */
+  launchAppSettings() {
+    return new Promise<boolean>((resolve, reject) => {
+      return this.executor.exec(
+        'launchAppSettings',
+        null,
+        response => resolve(MiniAppBridgeUtils.BooleanValue(response)),
+        error => reject(parseMiniAppError(error))
+      );
+    });
+  }
 }
--- a/js-miniapp-bridge/src/modules/utility-manager.ts
+++ b/js-miniapp-bridge/src/modules/utility-manager.ts
@@ -49,7 +49,7 @@ export class UtilityManager {
 
   /**
    * Trigger launchAppSettings from host
-   * @returns true or false whether launch app settings is enabled or not
+   * @returns true or false whether launch app settings is launched or not
    */
   launchAppSettings() {
     return new Promise<boolean>((resolve, reject) => {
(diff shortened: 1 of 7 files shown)


js-miniapp-bridge/src/modules/utility-manager.ts
```diff
@@ -46,4 +46,19 @@ export class UtilityManager {
       );
     });
   }
+
+  /**
+   * Trigger launchAppSettings from host
+   * @returns true or false whether launch app settings is enabled or not
```
rleojoseph, line 52:
```suggestion
   * @returns true or false whether launch app settings is launched or not
```

js-miniapp-sdk/src/miniapp.ts
```diff
@@ -308,4 +308,8 @@ export class MiniApp implements MiniAppFeatures, Ad, Platform {
   getPermissionStatus(name: PermissionName): Promise<string> {
     return getBridge().utilityManager.getPermissionStatus(name);
   }
+
+  launchAppSettings(): Promise<boolean> {
+    return getBridge().utilityManager.launchAppSettings();
+  }
```
rleojoseph, line 314: Do you think we can move this inside MiniAppUtils would be good ? Or new manager in SDK?
Because right now, if the MiniApps call MiniApp.launchSettings() would be kinda misleading you think ? Like its launching miniapp settings, but actually it is launching device settings.

js-miniapp-bridge/src/modules/utility-manager.ts
```diff
@@ -46,4 +46,19 @@ export class UtilityManager {
...
+   */
+  launchAppSettings() {
+    return new Promise<boolean>((resolve, reject) => {
+      return this.executor.exec(
+        'launchAppSettings',
+        null,
+        response => resolve(MiniAppBridgeUtils.BooleanValue(response)),
+        error => reject(parseMiniAppError(error))
+      );
+    });
```
rleojoseph, line 62: Thanks for the change.

We don't need to add the following in common-bridge.ts ?

  launchAppSettings() {
    return this.utilityManager. launchAppSettings();
  }

js-miniapp-sample/src/pages/launch-app-settings.js
```diff
@@ -0,0 +1,110 @@
...
+    );
+  });
+
+  function clear() {
+    setLaunchAppSettings('');
+  }
+
+  async function getLaunchAppSettings() {
+    try {
+      const result = await MiniApp.MiniAppUtils.launchAppSettings();
```
rleojoseph, line 79:
```suggestion
      const result = await MiniApp.miniappUtils.launchAppSettings();
```

